PR #24, opened 2021-09-09
Removed 'px'  and added on hover, active effect

voltrare:
removed 'px' only based on last commit
added hover effect on chapter, episode card

### DIFF
--- a/src/components/anime/EpisodeCard.tsx
+++ b/src/components/anime/EpisodeCard.tsx
@@ -43,7 +43,15 @@ const useStyles = makeStyles((theme) => ({
         marginRight: 16,
     },
     card: {
-        margin: '10px',
+        margin: 10,
+        '&:hover': {
+            backgroundColor: 'rgba(255, 255, 255, 0.2)',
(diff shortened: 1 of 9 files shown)


AriaMoradi:
> removed 'px' only based on last commit

why?

voltrare:
ah wait i am uploading the screenshot
dhul0ly7BV

> why?
ah because most of the css doesnt have 'px' so i thought this is to make it consistent
> Image?
only for the chaptercard and episode card

AriaMoradi:
The hover effect is really cool, a fine addition

> most of the css doesnt have 'px'

That's wrong, if it's missing the px, then the dev was lazy

AriaMoradi:
About the image it's for the style prop, not jss

voltrare:
ah ill change it back

AriaMoradi:
If you want make px units consistent then do it in a separate PR
Each PR should change one topic

voltrare:
allright then thanks for the insight

voltrare:
i should close this

PR #41, opened 2021-10-15
Better portrait support

minhe7735:
Image now fit by height when `viewport width > viewport height` and fit by width when `viewport width < viewport height`.

**Before**
before portrait
before landscape

**After**
after landscape
after landscape

### DIFF
--- a/src/components/manga/reader/Page.tsx
+++ b/src/components/manga/reader/Page.tsx
@@ -26,7 +26,7 @@ function imageStyle(settings: IReaderSettings): any {
         return () => {
             window.removeEventListener('resize', handleResize);
         };
-    });
+    }, []);
     if (settings.readerType === 'DoubleLTR'
     || settings.readerType === 'DoubleRTL'
     || settings.readerType === 'ContinuesHorizontalLTR'
@@ -47,8 +47,9 @@ function imageStyle(settings: IReaderSettings): any {
     return {
         display: 'block',
         marginBottom: settings.readerType === 'ContinuesVertical' ? '15px' : 0,
-        width: dimensions.width < dimensions.height ? '100vw' : 'auto',
-        height: dimensions.width < dimensions.height ? 'auto' : '100vh',
+        minWidth: '50vw',
+        width: dimensions.width < dimensions.height ? '100vw' : '100%',
+        maxWidth: '100%',
     };
 }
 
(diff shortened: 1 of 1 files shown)


src/components/manga/reader/Page.tsx
```diff
@@ -11,6 +11,22 @@ import SpinnerImage from 'components/SpinnerImage';
...
+            setDimensions({
+                height: window.innerHeight,
+                width: window.innerWidth,
+            });
+        }
+        window.addEventListener('resize', handleResize);
+        return () => {
+            window.removeEventListener('resize', handleResize);
+        };
+    });
```
AriaMoradi, line 29: follow this syntax for hook functions that have no dependency:

```
React.useEffect(() => {
    // callback body
}, []);
```

src/components/manga/reader/Page.tsx
```diff
@@ -31,9 +47,8 @@ function imageStyle(settings: IReaderSettings): any {
     return {
         display: 'block',
         marginBottom: settings.readerType === 'ContinuesVertical' ? '15px' : 0,
-        minWidth: '50vw',
-        width: '100%',
-        maxWidth: '100%',
+        width: dimensions.width < dimensions.height ? '100vw' : 'auto',
+        height: dimensions.width < dimensions.height ? 'auto' : '100vh',
```
AriaMoradi, line 51: setting min and max width prevents images with too small or too big width to be shown correctly, don't remove them and instead handle the new case with right values.

src/components/manga/reader/Page.tsx
```diff
@@ -105,7 +120,7 @@ const Page = React.forwardRef((props: IProps, ref: any) => {
     }, [handleVerticalScroll]);
 
     return (
-        <div ref={ref} style={{ margin: '0 auto' }}>
+        <div ref={ref} style={{ margin: 'auto' }}>
```
AriaMoradi, line 123: what's the effect of this change?

minhe7735: it more or less aligns the image to the center instead of being at top shown in the BEFORE portrait mode

minhe7735: **Single**
ltr


**Webtoon**
webtoon


**Continuous**
continous


**Computed for continuous and webtoon**
computed


**Computed for single**
computed single


I'm getting expected results when running it with Chrome/electron, center for single page and no margin for webtoon/continuous.

src/components/manga/reader/Page.tsx
minhe7735: re-added min/max width. Removed `height:` since it might stretch image due to `minWidth` and fitting by `height`

PR #56, opened 2021-10-30
added an aspect ratio to the manga card.

abhijeetChawla:
Added an aspect ratio to the MangaCard which prevents resizing of the component. Before this if an image failed to load the component would go in a Square shape if it alone in the row, If all the images in  a row failed to load then the whole row would change it's height,

### DIFF
--- a/src/components/MangaCard.tsx
+++ b/src/components/MangaCard.tsx
@@ -18,7 +18,10 @@ import SpinnerImage from 'components/util/SpinnerImage';
 const useStyles = makeStyles((theme) => ({
     root: {
         display: 'flex',
-        aspectRatio: '9/14',
+        // Preserve the shape of the manga cards, without this the manga card would chage it's
+        // if all the images in a row failed to load then the whole row would change into a
+        //  square shape
+        aspectRatio: '225/350',
     },
     wrapper: {
         position: 'relative',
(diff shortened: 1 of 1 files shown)


src/components/MangaCard.tsx
```diff
@@ -17,9 +17,8 @@ import SpinnerImage from 'components/util/SpinnerImage';
 
 const useStyles = makeStyles((theme) => ({
     root: {
-        height: '100%',
-        width: '100%',
         display: 'flex',
+        aspectRatio: '9/14',
```
AriaMoradi, line 21: how about the respected `225/350`? I know it's the same ratio but it's the golden number when making manga...
also add a comment for future generations reading the code.

AriaMoradi (changes requested):

src/components/MangaCard.tsx
```diff
@@ -17,9 +17,11 @@ import SpinnerImage from 'components/util/SpinnerImage';
 
 const useStyles = makeStyles((theme) => ({
     root: {
-        height: '100%',
-        width: '100%',
         display: 'flex',
+        // Preserve the shape of the manga cards, without this the manga card would chage it's
+        // if all the images in a row failed to load then the whole row would change into a
+        //  square shape
```
AriaMoradi, line 23: doesn't make sense

AriaMoradi (changes requested):

Issue #13, opened 2021-07-16
[Feature Request] Option to fit pages vertically on the screen


Miyazono1:
## What feature should be added to Tachidesk?
Add on option in the reader settings to fit pages on the screen vertically so you don't need to scroll to read them.

##  Why/Project's Benefit/Existing Problem
i personally think it's more convenient that way.

ntbm:
@AriaMoradi Want this as well. Will start working on it.

zFerry98:
I can't find the above option.
0.6.3 release.
![image](https://user-images.githubusercontent.com/17975351/177055272-40e198e5-5f16-4c95-94d1-8c88a881b1ca.png)

schroda:
done with https://github.com/Suwayomi/Tachidesk-WebUI/pull/323

PR #57, opened 2021-10-31
Add reader setting limiting page height to window size

ntbm:
Implements #13 
Added Setting in Reader for horizontal paginated Readers. It limits the page size to the window height in these cases.
Setting only shows up for valid reader selections

localhost_3000_manga_44_chapter_95(Small Laptop)
localhost_3000_manga_44_chapter_95(iPhone X)
localhost_3000_manga_44_chapter_95(Small Laptop) (1)
localhost_3000_manga_44_chapter_95(Small Laptop) (2)

### DIFF
--- a/src/components/navbar/ReaderNavBar.tsx
+++ b/src/components/navbar/ReaderNavBar.tsx
@@ -142,6 +142,7 @@ export const defaultReaderSettings = () => ({
     showPageNumber: true,
     continuesPageGap: false,
     loadNextonEnding: false,
+    fitToViewHeight: false,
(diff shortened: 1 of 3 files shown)


ntbm:
After reviewing and reverting https://github.com/Suwayomi/Tachidesk-WebUI/pull/41/files i come to the conclusion the my PR would just add more confusion to the current matter, so i'll pull it.
The PR mentioned above ist implementing a strech like behaviour on really large screens
Imho we need tachiyomi style scaling modes, so everyone knows what to expect in which mode.

PR #59, opened 2021-11-02
Added the Floating action button to Resume to the last read chapter

abhijeetChawla:
- When a chapter is marked as read/unread the lastPageRead is changed to 1,
- fixed a bug which made the boolean expression appear in the ui
- Migrated a file to mui v5 and also removed some unused Styles

### DIFF
--- a/src/screens/Manga.tsx
+++ b/src/screens/Manga.tsx
@@ -137,18 +137,19 @@ export default function Manga() {
         setFirstUnreadChapter(a);
     }, [chapters]);
 
-    const resumeButton = () => firstUnreadChapter && (
-        <Fab
-            sx={{ position: 'fixed', bottom: '5vw', right: '5vw' }}
-            component={Link}
-            variant="extended"
-            color="primary"
-            to={`/manga/${id}/chapter/${firstUnreadChapter.index}/page/${firstUnreadChapter.lastPageRead}`}
-        >
-            <PlayArrow />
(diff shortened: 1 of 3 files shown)


src/screens/Manga.tsx
```diff
@@ -158,7 +179,7 @@ export default function Manga() {
                     overscan={window.innerHeight * 0.5}
                 />
             </LoadingPlaceholder>
-
+            {resumeButton()}
```
ntbm, line 182: Create a Component here so it reads like <ResumeButton />

AriaMoradi (changes requested):

src/screens/Manga.tsx
```diff
@@ -158,7 +180,7 @@ export default function Manga() {
                     overscan={window.innerHeight * 0.5}
                 />
             </LoadingPlaceholder>
-
+            <ResumeButton />
```
ntbm, line 183:
```suggestion
            <ResumeFab />
```

ntbm (changes requested):

ntbm:
@AriaMoradi Are we now going with "exactly" where you left of aka `chapters.sort((a, b) => a.lastReadAt - b.lastReadAt)[0]`

AriaMoradi:
> @AriaMoradi Are we now going with "exactly" where you left of aka `chapters.sort((a, b) => a.lastReadAt - b.lastReadAt)[0]`

Can be in another PR

PR #61, opened 2021-11-05
Download filter and badge added

abhijeetChawla:
Added the download filter to the gallery
added the download badge to the manga card

### DIFF
--- a/src/util/useLibraryOptions.ts
+++ b/src/util/useLibraryOptions.ts
@@ -11,6 +11,8 @@ import { BooleanParam, useQueryParams, StringParam } from 'use-query-params';
 export type NullAndUndefined<T> = T | null | undefined;
 
 interface IUseLibraryOptions {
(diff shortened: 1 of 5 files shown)


src/util/useLibraryOptions.ts
```diff
@@ -33,6 +39,6 @@ export default function useLibraryOptions(): IUseLibraryOptions {
     // eslint-disable-next-line eqeqeq
     const active = !(unread == undefined);
```
ntbm, line 40: add the active condition here as well.

AriaMoradi:
why close?

src/util/useLibraryOptions.ts
abhijeetChawla: I will do that, but i am closing this pull request. I screwed up with git so i am closing this and making another pull request. I am so sorry

abhijeetChawla:
git screw up, i think

PR #76, opened 2021-11-14
load first page on read manga

### LINKED ISSUE #75: [Feature Request] Re-Reading A page
When you open a chapter of manga which you have already read, It takes you to the chapter you are currently reading. So if a person wants to re-read a chapter they can't do it because when they open the chapter they have already read it will take them to the last page of the chapter and because of the option [Load next chapter at ending] they will be moved to the chapter they are currently reading. So, In order to re-read a chapter they have to disable the [Load next chapter at ending] option and then when they tap on the chapter they have read, the last page of the manga will load and they have to manually go to the first chapter to re-read the chapter. 
If the chapter, that has been already read will load on the its first page it will be easier for people to re-read the chapter.

It will help people re-read chapter of manga fast.

AriaMoradi:
closes #75
PS: Do we want this behavior?

### DIFF
--- a/src/components/reader/pager/HorizontalPager.tsx
+++ b/src/components/reader/pager/HorizontalPager.tsx
@@ -87,8 +87,9 @@ export default function HorizontalPager(props: IReaderProps) {
     };
 
     useEffect(() => {
+        // scroll last read page into view after first mount
         pagesRef.current[curPage]?.scrollIntoView({ inline: 'center' });
-    }, [settings.readerType]);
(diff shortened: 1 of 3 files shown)


abhijeetChawla:
It Seems logical to open the first page when you have marked the chapter as read and it's same behaviour in Tachi

ntbm (approved):

ntbm:
+1 for this behaviour

PR #72, opened 2021-11-10
Mui 5 Migration of Some components

abhijeetChawla:
Migrated some components to the mui 5 system

PR #77, opened 2021-11-15
fixes mistakes from #72

AriaMoradi:
before:
image
after:
image


AriaMoradi:
There might be more, please test the app thoughtfully before approving.

abhijeetChawla:
I'll test some more now 😢

abhijeetChawla (commented):
I thought it looked nice with having some padding at the bottom for the rest of the app but The manga page looked wierd because of this though.

PR #78, opened 2021-11-15
Save tabs number in Url to persist tab when go to other paths

abhijeetChawla:
Adds an Optional parameter to the library path to provide which tab to go to. It also replaces the Tab Number in the url when tab is changed.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -158,7 +158,7 @@ export default function App() {
                                     <Route path="/manga/:id">
                                         <Manga />
                                     </Route>
-                                    <Route path="/library">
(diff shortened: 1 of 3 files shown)


ntbm (commented):
You need to handle not existing tabs and tabs which aren't a number as well

AriaMoradi (changes requested):
Two things are broken:
1- URLs like http://127.0.0.1:3000/library?unread=1 don't work
2- When a filter is active, changing tabs clears all filters

Maybe handling query prams is a better solution?

AriaMoradi (approved):

PR #88, opened 2021-11-15
migrate DoublePage to Mui 5

AriaMoradi:


ntbm (approved):
PR is fine, but we have a Bug here. The Double Pages can overflow with no way of doin a horizontal scroll to see the content
Bildschirmfoto 2021-11-30 um 19 10 35

AriaMoradi:
>  The Double Pages can overflow with no way of doin a horizontal scroll to see the content

Having that as an issue to fix later is desired.

PR #97, opened 2021-11-15
migrate SpinnerImage to Mui 5

AriaMoradi:


src/components/reader/Page.tsx
```diff
@@ -120,17 +103,26 @@ const Page = React.forwardRef((props: IProps, ref: any) => {
...
+
     return (
-        <div ref={ref} style={{ margin: 'auto' }}>
+        <Box ref={ref} sx={{ margin: 'auto' }}>
             <SpinnerImage
                 src={`${src}?useCache=${useCache}`}
                 onImageLoad={onImageLoad}
                 alt={`Page #${index}`}
                 imgRef={imgRef}
-                spinnerClassName={`${classes.image} ${classes.loadingImage}`}
```
AriaMoradi, line 130: Merging the two might have been wrong in the first place?

AriaMoradi:
semi-approved

ntbm (commented):
That's a complicated one.
The Page Spinner seems to be okay, but before the Page numbers are loaded the Spinner is window centered in the Old Version, while being at the same Position as the Page Spinner in the new Version

ntbm (approved):

AriaMoradi:
So does it need extra fixing?

ntbm (approved):
For me personally it's a non issue. The other loading indicator looks _nicer_. But replaced fast by the page spinner indicator so I approve.

AriaMoradi:
Lets sleep on this one a bit

src/components/reader/Page.tsx
abhijeetChawla: Probably yes, cause they both might have conflicting css. Having different styling objects for both makes more sense

ntbm (approved):
Looked at it again. Is fine for now considering the due rewrite of the reader.

PR #99, opened 2021-11-15
migrate Manga to Mui 5

AriaMoradi:


ntbm (approved):
We have a small change in the width of the Details Section here. Makes it a little wider. I personally like it. So approved.
Left new, right old
Bildschirmfoto 2021-11-17 um 12 21 19

PR #101, opened 2021-11-15
migrate SingleSearch to Mui 5

AriaMoradi:


AriaMoradi:
approved

ntbm (approved):
Layout change here. In my opinion for the better so i approve



Before:
Bildschirmfoto 2021-11-17 um 11 54 42


After:
Bildschirmfoto 2021-11-17 um 11 54 32

AriaMoradi:
> Layout change here. In my opinion for the better so i approve

It used to look like that before. V5 default styles are different so it had changed.

PR #110, opened 2021-12-03
Improves the Images in MangaCard

abhijeetChawla:
The Manga card Images don't Stretch or squish if the cover image of manga is of a different aspect ratio
previously 
Screen Shot 2021-12-03 at 11 00 22
This is how it looks now
Screen Shot 2021-12-03 at 11 00 09


AriaMoradi:
Does tachiyomi work similarly?

abhijeetChawla:
Yes Tachiyomi does the same thing

AriaMoradi (approved):

PR #124, opened 2022-01-18
Download Next Chapter while you read.

Silenc3IsGold3n:
This adds the feature of downloading the next chapter (if found) to the server as you read and adds a toggle-able UI setting for it in the Reader NavBar.  

image

This was done to decrease the time spent waiting for long chapters (like manhwa chapters) to load.


AriaMoradi:
Stale PR, closing.

PR #125, opened 2022-01-22
Refactor of ChapterList

abhijeetChawla:
Moved all the Data fetch logic for chapters to separate Hook called useChapterFetch

### DIFF
--- a/src/components/chapter/ChapterList.tsx
+++ b/src/components/chapter/ChapterList.tsx
@@ -12,11 +12,11 @@ import Typography from '@mui/material/Typography';
 import { CircularProgress, Fab } from '@mui/material';
 import { Link } from 'react-router-dom';
 import makeToast from 'components/util/Toast';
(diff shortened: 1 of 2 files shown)


src/components/chapter/ChapterList.tsx
```diff
@@ -191,6 +164,10 @@ export default function ChapterList(props: IProps) {
             </Fab>
         ));
 
+    if (noChaptersFound) {
+        makeToast('No chapters found', 'warning');
+    }
+
```
AriaMoradi, line 170: This will blast the toast on every render

AriaMoradi (changes requested):

src/components/chapter/ChapterList.tsx
abhijeetChawla: noChaptersFound is false by default so it would just check for the boolean on every render

Issue #28, opened 2021-09-14
[Feature Request] Hotkey support


BaccanoMob:
A few suggestions: 

Alt + arrows : move the page back or forward (or Esc to go back) [does not work in electron #27 ]

Ctrl+F : To move from browse/latest to search mode in extensions

Right click : to save page inside a manga (rather than searching inside folders) via a drop down menu possibly

Ctrl+O : to open a popup view of the library (or switch to library shortcut)

Ctrl+J : to open download queue

PR #129, opened 2022-02-18
Ctrl f overload

AriaMoradi:
This is old code and I don't know what state it's in, I won't look into it for now again.

contains some progress towards #28


AriaMoradi:
oops, duplicate of #116

PR #135, opened 2022-02-23
fix scrolling 8px

Robonau:
spoke about in discord
ChapterCount Header has a top margin of 8px that wasn't accounted for, leading to scrolling 8px on every chapter page
haven't tested


AriaMoradi:
Thanks!

PR #137, opened 2022-02-24
use withCredentials on requests

Robonau:
for those of us (me) that have cookie based authentication in front of the Tachidesk api and a WebUI on a different subdomain

is probably only relevant to me, so if you want to close rather than merge that's fine


Robonau:
nvm it doesn't work in some places

PR #138, opened 2022-02-24
more Download Queue info

Robonau:
add Imanga type definition to IDownloadChapter

add source name to queue
add manga name to queue

i have some permanently errored downloads and had no information visible other than what chapter it was

before:
Screenshot_20220224_190602
after:
Screenshot_20220224_190624

### DIFF
--- a/src/screens/DownloadQueue.tsx
+++ b/src/screens/DownloadQueue.tsx
@@ -125,7 +125,9 @@ export default function DownloadQueue() {
                                             </ListItemIcon>
                                             <ListItemText
                                                 primary={
-                                                    `${item.chapter.name} | `
+                                                    `${item.manga.source.name} | `
+                                                + `${item.manga.title} | `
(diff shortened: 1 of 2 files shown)


Robonau:
now:
Screenshot_20220225_182803

Robonau:
clicking it (anywhere other than the delete icon) takes you to the manga page now too

Robonau:
Screenshot_20220225_184150

Robonau:
Screenshot_20220225_184150

src/screens/DownloadQueue.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
AriaMoradi, line 1: fix this lint error, don't ignore it
Alternatively use styled for instead of sx if its not workable

AriaMoradi (changes requested):

PR #139, opened 2022-02-26
Genre filter

Robonau:
added genre filter for manga in library
genre options are generated from all the manga currently in library (on manga load)
alphabetical order

also separated activeSort and active (filter) (yellow when active)

old:
Screenshot_20220226_030814
new: (i couldn't get it to move in line with the check-boxes)
Screenshot_20220226_030846
clicking puts this on the right side of screen:
Screenshot_20220226_031021

no issues as far as i can see, was mentioned in discord.

### DIFF
--- a/src/components/library/LibraryMangaGrid.tsx
+++ b/src/components/library/LibraryMangaGrid.tsx
@@ -48,7 +48,7 @@ function genreFilter(queryY: NullAndUndefined<any[]>, queryN: NullAndUndefined<a
             ret = !queryN.some((v: string) => genre.includes(v));
         }
         if (queryY) {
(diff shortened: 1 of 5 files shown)


Robonau:
is now done with two arrays now rather than an object, seems to update faster now (less messy imo)
still cant get the arrays to be anything other than ```NullAndUndefined<any[]>``` throws an error if i try replace that `any` with anything else
though uses more url space though shouldn't be an issue

Robonau:
is broke rn

Robonau:
that fixed it

AriaMoradi:
Should we close this one?

Robonau:
cba to fix any of that
its not something that tachiyomi has
so imma just close it

PR #140, opened 2022-02-26
source genre filter

Robonau:
only on /sources/*/popular/ right now
much the same as the library version #139

(source equivalent to library files have been created)
add filter button on top right goes to the genre list (since there are no other filters already)
genre list  generated from the current manga list
screenshots same as #139

can be glitchy when scrolling with a filter, idk how to fix

### DIFF
--- a/src/components/source/SourceMangaGrid.tsx
+++ b/src/components/source/SourceMangaGrid.tsx
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+import React from 'react';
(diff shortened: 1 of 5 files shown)


Syer10:
I would personally deny this, sources are not garrentied to add the genres in browse, in fact its very rare for them to add any info other then the title and thumbnail. Adding this will only confuse people and make them think that its broken.

Robonau:
its something that the jui has and that ppl (at least 1) are asking for (in discord)

Robonau:
i should be able to make the button not show when there are no genres tho

Syer10:
JUI adds support for the source specific filters, not a custom filtering applied by the client.
Take a look here:
https://github.com/Suwayomi/Tachidesk-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/manga/controller/SourceController.kt#L65
And here:
https://github.com/Suwayomi/Tachidesk-Server/blob/master/server/src/main/kotlin/suwayomi/tachidesk/manga/impl/Search.kt#L38

Robonau:
ahh, lack of documentation really screwed me there

PR #142, opened 2022-03-01
Source filters, move search to SourceMangas

Robonau:
implementation of the `/api/v1/source/{sourceId}/filters` endpoint

Screenshot_20220301_222850
right side (mangadex, different per source):

Screenshot_20220301_222926

groups appear on top of the filters:
Screenshot_20220301_223010

something JUI already implemented


Robonau:
should i combine search in to browse
or should i move filtering over to search?
(personally i thing combining search in to browse makes more sense)
i currently have filter coded for the browse page (/popular), but filtering on the /search endpoint

Robonau:
rather than using resetFilterValue and updateFilterValue i should probably use a useState
and pass through the set

2. spamming too many options in quick succession will break it

AriaMoradi:
your commit naming is fun

Robonau:
i never know what to actually call my commits, so they tend to get pretty random

worked out the breaking that was happening in the last commit.
im probably using useState for more than is necessary, only really needed to do triggerUpdate and Data.
but im done for now, imma open for review.

Robonau:
screenshots update:
Screenshot_20220303_232253
Screenshot_20220303_232841
Screenshot_20220303_232853
Screenshot_20220303_232917

PR #156, opened 2022-03-06
in library badge to manga in sources

Robonau:
display when manga are already in library in sources
is something that tachiyomi does, just copying

compact and comfortable are basically the same
image


Robonau:
dont merge yet
i forgot about the other grid layouts

Robonau:
list:
Screenshot_20220306_220724

PR #157, opened 2022-03-07
mass search

Robonau:
add mass search
still slow, since it can only grab one source search at a time
for the record, i think this should be implemented server-side (with a WebSocket)

button to go to the search all page:
Screenshot_20220308_194726

(page is empty when first loading)
what the page looks like when you search:
Screenshot_20220308_194847

when the source cant find anything:
Screenshot_20220308_194957

https://github.com/Suwayomi/Tachidesk-Server/issues/209
though that is a server issue


Robonau:
il need to fix this after #156 goes through, so its staying a draft for now

Robonau:
concurrency makes it quite a bit faster, but its still not as fast as tachiyomi
perhaps im just using slow sources
https://streamable.com/ejxh1g

AriaMoradi:
remove `package-lock.json` the project uses yarn for managing packages.

AriaMoradi:
something might be wrong with your `yarn.lock` all the links have changed

Robonau:
should be fine now

PR #165, opened 2022-03-24
items per row setting

Robonau:
adds setting to change the manga per rows (for compact and comfortable grid)
4 different settings cause 4 different widths catered for
idk what to name them for users to understand (need help) so am currently just using the codenames
(tachiyomi just has landscape and portrait so not applicable)
Screenshot_20220324_003948
Screenshot_20220324_004000
lg = 10 example:
Screenshot_20220324_004041

### DIFF
--- a/src/components/MangaCard.tsx
+++ b/src/components/MangaCard.tsx
@@ -71,6 +71,7 @@ const truncateText = (str: string, maxLength: number) => {
 interface IProps {
     manga: IMangaCard
     gridLayout: number | undefined
+    dimensions: number
 }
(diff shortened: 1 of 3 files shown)


AriaMoradi:
this needs a more user friendly solution

Robonau:
> this needs a more user friendly solution

thats why i was asking for help

Robonau:
i copied how jui deals with the manga grid
(try keep every item close to a given width whilst still filling the width)
Screenshot_20220328_211009
image
slider is from 100-1000(px) (i doubt anyone will want anything outside of this, but the number input is there just encase)

PR #168, opened 2022-04-04
unified library options

infix:
Changes:
1. renamed `LibraryDisplayOptions` to `LibraryOptions`, and updated it to include filter options
2. updated `LibraryOptionsContext`
3. removed `useLibraryOptions`

so the rational behind this pr is that every time a user clicks on the library tab button it resets all the select filters (downloaded, unread). that results in them having to re-add the filters.

options are now stored in the `localStorage` instead of the query-params, and unifies the way that options are handled so that there's not a lot of duplication. this brings it more in line with how tachiyomi works


Robonau:
seems good

PR #177, opened 2022-06-22
use displayValues for select filters

Robonau:
this will fix mangadex filters after tachiyomi changed formats
(will only work for server versions after Syer10's patch to server Suwayomi/Tachidesk-Server#347)

### DIFF
--- a/src/components/source/SourceOptions.tsx
+++ b/src/components/source/SourceOptions.tsx
@@ -88,7 +88,7 @@ export function Options({
                             <SelectFilter
                                 key={`filters ${e.filter.name}`}
                                 name={e.filter.name}
-                                values={e.filter.values}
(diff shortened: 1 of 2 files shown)


Syer10:
You could check if displayValues is present/null and use whatever values returns if it is. In JUI I show the raw json objects in the array if displayValues is null

Robonau:
> You could check if displayValues is present/null and use whatever values returns if it is. In JUI I show the raw json objects in the values array if displayValues is null

if it's not a string it will error and crash
making it not crash is allot more work

AriaMoradi:
> (will only work for server versions after Syer10's patch to server [Suwayomi/Tachidesk-Server#347](https://github.com/Suwayomi/Tachidesk-Server/pull/347))

Is that not in the latest Server release?

PR #179, opened 2022-07-11
Replace current history instead of pushing on search search

intrnl:
Currently, each keystroke being made on search would result in a new history entry being pushed, which is unideal, because it means that the back/forward entries are filled up with unnecessary entries. This also affects the app's own back button that is present in some screens, the button would just be undoing what you've typed in search earlier instead of actually going to the previous screen, which is more ideal.

This change makes it so that the current history entry is being replaced instead, using `replaceIn` instead of the default `pushIn` [update type](https://github.com/pbeshai/use-query-params/blob/master/packages/use-query-params/README.md#urlupdatetype).

### DIFF
--- a/src/components/util/AppbarSearch.tsx
+++ b/src/components/util/AppbarSearch.tsx
@@ -27,10 +27,10 @@ const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
     const inputRef = React.useRef<HTMLInputElement>();
 
     function handleChange(e: React.ChangeEvent<HTMLInputElement>) {
-        setQuery(e.target.value === '' ? undefined : e.target.value);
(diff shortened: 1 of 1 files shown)


intrnl:
While Safari's limit is 100 changes/30 sec before it would throw if you make any more pushState/replaceState calls, personally it seems like it would be better to just defer committing the query params entirely.

Robonau:
yea, i wanted to have it work on press enter but at the time ArMor (lead dev) wouldn't let me

AriaMoradi:
Stale PR, closing.

intrnl:
this can still be merged, no?

intrnl:
@AriaMoradi this can still be taken care of, no?

AriaMoradi:
Sure

Issue #123, opened 2022-01-08
[Feature Request] Multiple-select or select option for downloading chapters


thisisnaveen:
It will be easier to download if there is a multiple-select option or a select option to download multiple chapters. Kindly, consider it. Or else add the same feature for download as mark previous download like in mark previous as read feature.

martinek:
@AriaMoradi This is now also possible with #194

Issue #38, opened 2021-09-21
selecting multiple/bulk action for chapters


smarttommyau:
selecct chapters to read,unread,bookmark and download

martinek:
@AriaMoradi This was fixed with #194

PR #191, opened 2022-11-07
Feature/batch chapter download

martinek:
There is lot of changes in this one. I had to do some cleanup to make room for chapter selection so keep with me please.

- Chapter components were moved to manga directory, since they are only used in manga screen.
- There is now a React way to render actions through portal into toolbar (see `NavbarToolbar` component). This is better then setting actions through `useEffect` as it automatically cleans up after itself and is much nicer to use.
- Code around using chapterOptions was moved into `util.tsx` and is now exposed through single `useChapterOptions`.
- Manga categories was changed 

Tachidesk_and_Tachidesk

- ChapterOptions button was moved from Chapters list to manga toolbar.
- ChapterOptions were cleaned up. Mostly paddings and alignment was fixed. I have also changed rendering of Sort tab options to look cleaner and fixed missing keys.

image
image
image

- Refresh action and category edit action was moved to "More" menu of manga. While data is refreshing, small progress icon is shown in toolbar.
image
image

- ChapterCard layout was changed little bit.
- Date and download status are now on separate row so that they are in the same area on each item.
- Bookmark icon is now primary color and correctly aligned

image

- While download is running, progress bar is visible at the bottom of ChapterCard

image

- Chapter menu was cleaned up and icons were added to each action. Actions behave the same as before.
image
image

Finally changes that have to do with selection :)

- When "Select" action is clicked chapter is selected and UI changes to "selection mode"

image

While in selection mode

- Whole chapter card behaves like button for checkbox.
- User can select and deselect chapters.
- Floating action button at the bottom changes to "Selection FAB".
- Number of selected chapters is shown on the FAB.
- When all chapters are deselected, selection mode is canceled.
- There is button for Select All which adds all visible chapters to selection.
- There is button for Clearing the selection. This practically closes selection mode.
- Selection FAB currently has only single action "Download". This adds selected chapters to download queue.

image


martinek:
Forgot to link to issues requesting this...
- #123 
- partially #38

AriaMoradi:
Another huge PR :/

PR #192, opened 2022-11-10
Memoize empty view face so it does not change on rerender

martinek:
This fixes small issue with EmptyView

Currently when EmptyView rerenders, the random face is always regenerated. Since EmptyView can be rerendered multiple times, this causes faces to change and looks weird.

This PR memoizes the random face, so once EmptyView renders it does not change the face.


martinek:
Wrong branch, my bad :P

PR #200, opened 2022-11-24
Verticall scroll navigation and fix

### LINKED ISSUE #14: [Bug] If the last page is not the only page visible it never gets counted as read
## Device information
- Tachidesk version: stable v0.4.3
- Server Operating System: Ubuntu 20.04.02 / docker
- Server JVM version: docker
- Client Operating System: win10
- Client Web Browser: Firefox 89.0.2

## Steps to reproduce
1. Open a manga
2. Read to the last page
3. If the last page is not the only page visible it never gets counted as read and the chapter as well

### Expected behavior
Mark last page as read as soon as it is visible.

### Actual behavior
Last page is not marked as read even if fully visible, as long as the second to last page is still visible too.

## Other details
You can resize the browser window so the last page is the only visible, then it gets marked as read.

### Example

Works:
grafik

Doesn't work:
grafik

### LINKED ISSUE #162: [Bug]  Space/Tap/Click should Scroll down the page instead of going to next image
## Device information
- Tachidesk version: v0.6.2-r1074-linux-x64
- Server JVM version: OpenJDK 8u312
- Client Operating System: Same or Android
- Client Web Browser: Chrome/Firefox

## Steps to reproduce
1. Open a comic with large images
2. Make it so that you'll need to scroll down to see the full image (or just open a webtoon, they have long pages)
3. Click space or tap/click on the "next" area.

### Expected behavior
Space/Tap/Click should Scroll down the page instead of going to next image

### Actual behavior
Space/Tap/Click will take you to the next image regardless of how much you actually read/viewed.

## Other details
This behavior is more apparent when reading in "Webtoon" or "Continuous vertical" mode, and especially with webcomics/webtoons as they have very long images.

martinek:
This PR changes the scrolling behavior in `Continues vertical` and `Webtoon` layouts. Instead of going to "next page", when user taps / clicks / preses key the view now scrolls 95% of its height. This makes reading webtoons much better. This fixes #162. 

I have not done the complete "magic" behavior, as described in the issue. I think the "if page is visible, scroll to the top of it" (snapping) would be weird for webtoons. I tried it on some "same page size" manga and the experience is very dependant on screen size.

In future the behavior could be configurable so user can specify how much it should scroll and also if the scroll should be smooth or instant. We could also add option for the "snapping" behavior described above.

While doing this, I have refactored some behavior in both vertical and horizontal scrolling layouts and added fix for different issue. This also fixes #14. When view is scrolled all the way (either direction) the last page number is now sent as current.

### DIFF
--- a/src/components/reader/Page.tsx
+++ b/src/components/reader/Page.tsx
@@ -5,7 +5,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import React, { useEffect, useRef } from 'react';
+import React, { useRef } from 'react';
 import SpinnerImage from 'components/util/SpinnerImage';
 import useLocalStorage from 'util/useLocalStorage';
 import Box from '@mui/system/Box';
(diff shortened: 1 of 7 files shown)


alkumaish:
Tested and it works great.  But Shift+space should go to prev page (à la PgUp)   right now the shift key is ignored.
This is not really a bug because everything works as intended, It's just for the sake of consistency (to make webUI's behave like normal scrollable pages).

PR #216, opened 2023-01-01
Feature/#213 reader use and save reader settings per manga

### LINKED ISSUE #213: [Feature Request] Save reader settings per manga
## What feature should be added to Tachidesk?
Each manga should have their own reader settings instead of using the same reader settings for every manga.

Currently there are only global reader settings.
By having custom reader settings for each manga you don't have to update the settings each time you switch to a different manga that you want to read with different reader settings.

E.g.:
Switching from a webtoon to a manga. (webtoon: Reader Type = Webtoon; manga: Reader Type = Duoble Page (LTR)

schroda:
I know quite big but e.g. the "meta" property wasn't added yet and adding the logic to use specific reader settings for each manga uncovered some bugs in the "ReadNavBar"

adds:
 - logic to be able to use the "meta" property
   - global server
   - manga
   - chapter
   - category
 - save and load/use the reader settings for each manga

fixes:
 - ReadNavBar (required for the new logic to work)
   - opening the navigation bar when passed props changed (when necessary)
   - hiding/showing the "openButton"
     - on first render
     - when clicking the button
   - keeping the state when opening the prev/next chapter
 - Reader
   - only request the manga data again in case the "mangaId" changed (instead of basing it on the "chapterIndex"

Closes #213


AriaMoradi:
Dude, slow down!

schroda:
> Dude, slow down!

that was it (well, + the 2 other prs 😆)

Issue #214, opened 2023-01-01
[Feature Request] The default reader settings should be changable


schroda:
## What feature should be added to Tachidesk?
The default reader settings used for mangas that haven't been read yet should be changeable in the settings.

##  Why/Project's Benefit/Existing Problem
In case you use the same reader settings most of the time and are only changing them for some mangas, it is annoying to always have to change the "default" used reader settings

PR #218, opened 2023-01-01
Feature/#215 add logic for metadata migration

### LINKED ISSUE #215: [Feature Request] As a developer I want to be able to change metadata keys without the users losing their data
## What feature should be added to Tachidesk?
As a developer I want to be able to change metadata keys (the keys saved in the "meta" property of e.g. the manga) without having users lose their data.

It should be possible to improve metadata key names (e.g. they have a typo or a different name would be more suiteable) without having users lose their data that has been saved under the old metadata keys.

schroda:
Closes #215

- wait for #214 to get merged
- rebase onto master


schroda:
rebased onto #214

PR #223, opened 2023-01-09
Feature/force absolute import path

schroda:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
         'airbnb',
         'airbnb-typescript'
     ],
-    plugins: ['@typescript-eslint'],
+    plugins: ['@typescript-eslint', 'no-relative-import-paths'],
     parserOptions: {
         project: './tsconfig.json',
     },
@@ -24,5 +24,13 @@ module.exports = {
         'react/jsx-no-bind' : 'off',
         'react/jsx-props-no-spreading': 'off',
         'react/require-default-props': 'off',
+
+        "no-relative-import-paths/no-relative-import-paths": [
+            "error",
+            {
+                rootDir: "src",
+                prefix: false,
+            },
+        ],
     },
 };
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "eslint-config-airbnb-typescript": "^14.0.0",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-no-relative-import-paths": "^1.5.2",
     "eslint-plugin-react": "^7.21.5",
     "eslint-plugin-react-hooks": "^1.7.0",
     "typescript": "^4.8.4"
--- a/src/components/MangaGrid.tsx
+++ b/src/components/MangaGrid.tsx
@@ -13,8 +13,8 @@ import EmptyView from 'components/util/EmptyView';
 import LoadingPlaceholder from 'components/util/LoadingPlaceholder';
 import { Typography } from '@mui/material';
 import { Box } from '@mui/system';
-import MangaCard from './MangaCard';
-import { GridLayout } from './context/LibraryOptionsContext';
+import MangaCard from 'components/MangaCard';
+import { GridLayout } from 'components/context/LibraryOptionsContext';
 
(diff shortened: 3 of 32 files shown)


AriaMoradi (changes requested):
I don't like this one in particular, replace `src` with `@`

schroda:
> I don't like this one in particular, replace `src` with `@`

what about `@webUI`?

AriaMoradi:
> > I don't like this one in particular, replace `src` with `@`
> 
> what about `@webUI`?

why? we don't have some other root than `src`?

schroda:
> > > I don't like this one in particular, replace `src` with `@`
> > 
> > 
> > what about `@webUI`?
> 
> why? we don't have some other root than `src`?

I guess I'm missing something, what exactly do you mean with replace `src` with `@`?
(1.) `src/component/...` -> `@/component/...`
or
(2.) `src/component/...` -> `@component/...`

(1.) doesn't work, would have to be `@<XYZ>/components...`


but ignoring all that, it looks like create-react-app doesn't support alias imports anymore (tsconfig -> "paths"...).

> yarn start
yarn run v1.22.19
$ react-scripts start
The following changes are being made to your tsconfig.json file:
>  - compilerOptions.paths must not be set (aliased imports are not supported)

I don't really know how else I could do it?

AriaMoradi:
take a look: https://github.com/ixartz/Next-js-Boilerplate/blob/81e3f7160b8ec977526427d25ea8f7dcb60ee2c2/tsconfig.json#L40

AriaMoradi:
> looks like create-react-app doesn't support alias imports anymore
> 
> > Now that is a problem

Compare
```jsx
import MyComponent from '@/components/MyComponent';
```
vs
```jsx
import MyComponent from 'src/components/MyComponent';
```
- The first one has a clear visual indicator that we are importing local code, the second not so much.
- That's the syntax I like to see anyway, if CRA doesn't want to support it then either give up or find an alternative solution?

Looks like I have to reject this? why would we even bother making absolute imports and starting from `src` mandatory? It's just more typing...

schroda:
> That's the syntax I like to see anyway, if CRA doesn't want to support it then either give up or find an alternative solution?
ye then I guess we leave it as it is

> The first one has a clear visual indicator that we are importing local code, the second not so much.
I get what you mean but since it's not as straight forwards to do it as I thought I'm not going to invest more time in reasearching

I'd say doing it with `src` is closer to your liked solution (`@`)  then how it's currently done (`component/...` or `screen/...`) since there is no visual identification that it's from the same source

> why would we even bother making absolute imports and starting from src mandatory? It's just more typing...

absolute imports: `CONTRIBUTING.md` -> `Coding Style Guide`:
> Don't use relative imports.


> It's just more typing...

I mean the IDE is doing everything for you


but in the end it's your call if you prefer the way it's done currently you can just reject it of course

AriaMoradi:
considering everything and that we set a rule to not have absolute imports, if something like `@src/` works then it shall be so, but if we can't include `@` then let's just have the no relative import lint rule and keep the `src/` part out as it's redundant path your eye doesn't need to see, on the other hand, at least one can make some sense in their mind `components/..` means we are using a component, that some beneficial visual indication if I say so myself.

schroda:
Alright, then it's the no relative import rule without `src` since alias imports do not work at all with CRA, gets removed from the tsconfig once you run a script (e.g. `start` or `build`)


----

`@/...` would work without CRA, it was just an issue with the eslint settings from `airbnb` (`ESLint: Missing file extension for "@/util/client"(import/extensions)`)

regarding `@/...`, `src/...` or `components/...` I guess we agree to disagree 😆 
me personally idc if it would be `@` or `src`, I don't really see a difference there, but regarding `components/...` I'd say it's not really clear were it's coming from, ofc you can tell that it's components that you are using, but you can't really tell if it's local or from some library.
I mean e.g. you could have a local folder named exactly like a library that you're using and then it would be possible to have an import for the library and for files from the local folder, which would just be confusing.

AriaMoradi:
> I mean e.g. you could have a local folder named exactly like a library that you're using and then it would be possible to have an import for the library and for files from the local folder, which would just be confusing.

In that case I would agree with your point, but we don't have such issue with this code base.

AriaMoradi (approved):

PR #227, opened 2023-02-05
Add logic to migrate metadata values

schroda:
Makes it possible to rename or fix values of already saved metadata without the user losing his data


martinek (approved):
This looks fine, for what it is supposed to do.

PR #232, opened 2023-02-05
Add GitHub Action to run tsc on push

schroda:

### DIFF
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -1,11 +1,17 @@
-name: Lint
-on: push
+name: CI Pull Request
+on:
+  pull_request:
 jobs:
   build:
+    name: ci-pull-request
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Install modules
         run: yarn
+
       - name: Run ESLint
         run: yarn lint
+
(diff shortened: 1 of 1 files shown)


AriaMoradi:
Also add this to pull request checks.

schroda:
you can make the action "code-checks" required to pass for merging to be allowed (branch protection rule -> require status cehcks to pass before merging)

AriaMoradi:
> Also add this to pull request checks.

What I am reffering to is
https://github.com/Suwayomi/Tachidesk-Server/blob/master/.github/workflows/build_pull_request.yml

Look at the directory as well.

schroda:
> It should run everything we do for push up until making a release.

currently it runs on every push, so I think adding "pull_request" doesn't really add anything, unless it should only run for pull requests and ignore other branches (so replacing "push" with "pull_request")?


> Look at the directory as well.

not sure what you mean


> you can make the action "code-checks" required to pass for merging to be allowed (branch protection rule -> require status checks to pass before merging)

I think adding this after the pr is merged would still make sense just to make it a little safer

schroda:
I think this should be fine now

AriaMoradi:
@akabhirav review please

akabhirav (approved):
LGTM

PR #238, opened 2023-02-12
Adds search by genre to WebUI

akabhirav:
# What
As the title says, this feature will allow you to use the search input with genre as well. You can input comma separated genre to search by them
**Note: You need to put the full genre text (case does not matter)**
# Why
Searching by genre is very cool, it has the ability to give you exactly what you are in the mood for. Want something that has a lot of action in it, but also monsters. Type in `action, monster` and it will give you all the mangas in your library that are tagged with them
**Disclaimer: Search by genre is subject to information provided by extension and sources. So if your source/extension does not provide proper genre details, this will not work**

### DIFF
--- a/src/components/library/LibraryMangaGrid.tsx
+++ b/src/components/library/LibraryMangaGrid.tsx
@@ -53,6 +53,7 @@ const filterManga = (
     unread: NullAndUndefined<boolean>,
     downloaded: NullAndUndefined<boolean>,
 ): IMangaCard[] => {
+    let filteredManga: IMangaCard[] = [];
     if (query) {
         const titleFilteredManga = manga.filter((m) => queryFilter(query, m));
         const genreFilteredManga = manga.filter((m) => queryGenreFilter(query, m));
@@ -63,15 +64,12 @@ const filterManga = (
             }
             return acc;
         }, {});
-        return Object.values(unique);
+        filteredManga = Object.values(unique);
     }
-    return manga.filter((m) => {
-        if (query) {
-            return queryFilter(query, m);
-        }
-
-        return downloadedFilter(downloaded, m) && unreadFilter(unread, m);
-    });
+    filteredManga = (filteredManga.length ? filteredManga : manga).filter(
+        (m) => downloadedFilter(downloaded, m) && unreadFilter(unread, m),
+    );
+    return filteredManga;
 };
 
 const sortByUnread = (a: IMangaCard, b: IMangaCard): number =>
--- a/src/components/library/LibraryMangaGrid.tsx
+++ b/src/components/library/LibraryMangaGrid.tsx
@@ -41,19 +41,38 @@ const queryFilter = (query: NullAndUndefined<string>, { title }: IMangaCard): bo
     return title.toLowerCase().includes(query.toLowerCase());
 };
 
+const queryGenreFilter = (query: NullAndUndefined<string>, { genre }: IMangaCard): boolean => {
+    if (!query) return true;
+    const queries = query.split(',').map((str) => str.toLowerCase().trim());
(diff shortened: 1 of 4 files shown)


src/components/library/LibraryMangaGrid.tsx
```diff
@@ -41,19 +41,36 @@ const queryFilter = (query: NullAndUndefined<string>, { title }: IMangaCard): bo
...
+    if (query) {
+        const titleFilteredManga = manga.filter((m) => queryFilter(query, m));
+        const genreFilteredManga = manga.filter((m) => queryGenreFilter(query, m));
+        const unique = titleFilteredManga.concat(genreFilteredManga).reduce((acc: Record<number, IMangaCard>, obj) => {
+            const { id } = obj;
+            if (!acc[id]) {
+                acc[id] = obj;
+            }
+            return acc;
+        }, {});
```
akabhirav, line 66: This will keep the mangas that matched by title above the ones that matched by genre. I am working on a more streamlined version that gives a rank to manga objects while filtering, and sorts accordingly. Could be more performant

src/components/library/LibraryMangaGrid.tsx
```diff
@@ -63,15 +64,12 @@ const filterManga = (
             }
             return acc;
         }, {});
-        return Object.values(unique);
+        filteredManga = Object.values(unique);
```
schroda, line 67: this commit reverts #226.
not sure how everyone else sees this, but I'd like to be able to search my whole library and not just the filterted one (e.g. downloaded, unread), since otherwise, if you want to open a manga that is not included in the filters, you have to unset the filters everytime you search for something like that (and set them again after you've found the manga)

akabhirav: Is this not how the app works? If you want to search something that is not downloaded then you remove the filter and search for it, right?

It felt like a bug that my filters were not being honored when I searched for my manga

I believe we should stick close to the app implementation. I think we should not deviate too far from the tachiyomi app. If setting and unsetting the filters are cumbersome or buggy we could improve the experience there.  I will adhere to the consensus. But, I think we should keep this.

schroda: I never used the app (don't have an android), so I can't say.

I guess it's preference. I expected it to search my whole library (or at least the category, since whole library is not possible).
since I know what I'm searching for (a specific title), I wouldn't want a filter to mess with it.

if it should work like the app, what about adding some kind of flag to search all mangas/ignore filters?

akabhirav: That sounds like a good workaround, but for that we need to clarify the following details
1. Whether this preference will be stored in localStorage or sever
2. Where can users find this preference(Settings page or somewhere else)
3. Will it be on/off by default
4. Any other solutions like displaying a fab while searching and if filters are active, to quicky disable the filters(either for that search or permanently)
5. Should it be done in this PR or another(I prefer another PR)

akabhirav: @schroda any opinions on these? I'd like to get this PR reviewed

schroda: 1. I'd say server as well
2. maybe the library "settings" (filter, sort, ui), or a settings button inside the search, which is only visible when search is open?
3. should be the same as the app then, so disabled
4. I wouldn't mess with the filters, since it's also not really required for it to work
5. yes, should be done in another pr

akabhirav: @schroda I have created a new PR https://github.com/Suwayomi/Tachidesk-WebUI/pull/242 to find a solution for the issue in this thread

schroda (approved):

PR #242, opened 2023-02-15
Introduced override filters while searching setting

akabhirav:
# What
This PR introduces a new setting called `Ignre Filters while Searching`. This setting as it says determines whether manga search will respect the filters that are set such as Downloaded, Unread, etc.

# Why
This will resolve a previous issue that some users had with the search behaviour which resulted in #226. This PR completely ignores the filters while searching. This PR is middle ground to keep most users satisfied with the Web UI behaviour

### DIFF
--- a/src/components/library/LibraryMangaGrid.tsx
+++ b/src/components/library/LibraryMangaGrid.tsx
@@ -11,6 +11,7 @@ import MangaGrid from 'components/MangaGrid';
 import { useLibraryOptionsContext } from 'components/context/LibraryOptionsContext';
 import { StringParam, useQueryParam } from 'use-query-params';
 import { useMediaQuery, useTheme } from '@mui/material';
+import { useSearchSettings } from 'util/searchSettings';
 
 const FILTERED_OUT_MESSAGE = 'There are no Manga matching this filter';
 
@@ -46,13 +47,16 @@ const filterManga = (
     query: NullAndUndefined<string>,
     unread: NullAndUndefined<boolean>,
     downloaded: NullAndUndefined<boolean>,
+    overrideFilters: boolean,
 ): IMangaCard[] =>
     manga.filter((m) => {
+        const isFiltered = downloadedFilter(downloaded, m) && unreadFilter(unread, m);
         if (query) {
-            return queryFilter(query, m);
+            const isQueried = queryFilter(query, m);
+            if (overrideFilters) return isQueried;
+            return isFiltered && isQueried;
         }
-
-        return downloadedFilter(downloaded, m) && unreadFilter(unread, m);
+        return isFiltered;
     });
 
 const sortByUnread = (a: IMangaCard, b: IMangaCard): number =>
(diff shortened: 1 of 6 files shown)


src/screens/settings/SearchSettings.tsx
```diff
@@ -0,0 +1,33 @@
...
+import makeToast from 'components/util/Toast';
+import ListItemIcon from '@mui/material/ListItemIcon';
+import SearchIcon from '@mui/icons-material/Search';
+
+export default function SearchSettings() {
+    const { metadata, settings } = useSearchSettings();
+
+    const setSettingValue = (key: keyof ISearchSettings, value: boolean) => {
+        requestUpdateServerMetadata(metadata ?? {}, [[key, value]]).catch(() =>
+            makeToast('Failed to save the default reader settings to the server', 'warning'),
```
schroda, line 15: you forgot to update the error message (still says reader settings)

src/screens/settings/SearchSettings.tsx
```diff
@@ -0,0 +1,33 @@
...
+        );
+    };
+    return (
+        <ListItem>
+            <ListItemIcon>
+                <SearchIcon />
+            </ListItemIcon>
+            <ListItemText primary="Override Filters when Searching" />
```
schroda, line 23: what do you think about something like "Ignore filters while searching"? I feel like that is easier to understand then "Override"

src/util/searchSettings.ts
```diff
@@ -0,0 +1,30 @@
+import { useQuery } from 'util/client';
+import { getMetadataFrom } from 'util/metadata';
+
+export const getDefaultSettings = () =>
+    ({
+        overrideFilters: false,
+    } as ISearchSettings);
+
+const getReaderSettingsWithDefaultValueFallback = (
```
schroda, line 9: same as for the error message, name inlcudes "ReaderSettings"

src/components/library/LibraryMangaGrid.tsx
```diff
@@ -53,12 +54,12 @@ const filterManga = (
     query: NullAndUndefined<string>,
     unread: NullAndUndefined<boolean>,
     downloaded: NullAndUndefined<boolean>,
+    overrideFilters: boolean,
```
schroda, line 57: I think "ignoreFilters" would be better fitting over all.
"overrideFilters" sounds to me like the filters will get replaced with different ones, while they are just ignored if the flag is set to true

src/typings.ts
```diff
@@ -81,7 +81,7 @@ export interface IMetadataHolder<VALUES extends AllowedMetadataValueTypes = stri
 
 export type AllowedMetadataValueTypes = string | boolean | number | undefined;
 
-export type MangaMetadataKeys = keyof IReaderSettings;
+export type MangaMetadataKeys = keyof (IReaderSettings & ISearchSettings);
```
schroda, line 84: AppMetadataKeys should be a better place to add the type, since the search settings aren't really part of the mangas

schroda (approved):
LGTM

PR #246, opened 2023-02-19
Feature/add translation keys

schroda:


src/components/SourceCard.tsx
```diff
@@ -120,7 +123,7 @@ const SourceCard: React.FC<IProps> = (props: IProps) => {
                                 </Button>
                             )}
                             <Button component={Link} to={`/sources/${id}/popular/`} variant="outlined">
-                                Browse
+                                {t('SourceBard.button.browse')}
```
AriaMoradi, line 126: ScoreCard

src/components/library/LibraryMangaGrid.tsx
```diff
@@ -11,8 +11,9 @@ import MangaGrid from 'components/MangaGrid';
 import { useLibraryOptionsContext } from 'components/context/LibraryOptionsContext';
 import { StringParam, useQueryParam } from 'use-query-params';
 import { useMediaQuery, useTheme } from '@mui/material';
+import { useTranslation } from 'react-i18next';
 
-const FILTERED_OUT_MESSAGE = 'There are no Manga matching this filter';
+const FILTERED_OUT_MESSAGE = 'LibraryMangaGrid.filter.label.no_matches';
```
AriaMoradi, line 16: t('LibraryMangaGrid.filter.label.no_matches')

import t from i18next directly

src/components/library/LibraryOptionsPanel.tsx
```diff
@@ -13,11 +13,12 @@ import ThreeStateCheckboxInput from 'components/atoms/ThreeStateCheckboxInput';
...
 import React from 'react';
+import { useTranslation } from 'react-i18next';
 
 const TITLES = {
-    filter: 'Filter',
-    sort: 'Sort',
-    display: 'Display',
+    filter: 'LibraryOptionsPanel.title.filter',
+    sort: 'LibraryOptionsPanel.title.sort',
+    display: 'LibraryOptionsPanel.title.display',
```
AriaMoradi, line 21: use t function when defining these

src/i18n/locale/en.json
```diff
@@ -14,6 +14,297 @@
...
+    "SourceMangas": {
+      "no_mangas_found": "No manga was found!",
+      "local_source": {
+        "label": {
+          "checkout": "Check out",
+          "guide": "Local source guide"
+        }
+      }
+    },
+    "settings": {
```
AriaMoradi, line 27: be consistent in capitalization

AriaMoradi (changes requested):

src/components/library/LibraryMangaGrid.tsx
schroda: that only works in case the langauge is fixed and can't be changed, otherwise, in case another language gets selected after the first load, the language would not get updated

and atleast I don't want to have to use my native language, which gets selected by default

src/components/library/LibraryOptionsPanel.tsx
schroda: [same](https://github.com/Suwayomi/Tachidesk-WebUI/pull/246#discussion_r1113623031)

src/i18n/locale/en.json
schroda: thats the sub folder inside "screen"
do you think it would be better to omit it and just have the screens inside it be placed directly in "screens"?

AriaMoradi: how about `screens.settings.`

schroda: thats the current key

PR #251, opened 2023-02-25
Show empty library in case search doesn't match anything

schroda:
In case the search didn't match anything all mangas where shown instead of an empty library.

### DIFF
--- a/src/components/library/LibraryMangaGrid.tsx
+++ b/src/components/library/LibraryMangaGrid.tsx
@@ -53,25 +53,13 @@ const filterManga = (
     query: NullAndUndefined<string>,
     unread: NullAndUndefined<boolean>,
     downloaded: NullAndUndefined<boolean>,
-): IMangaCard[] => {
-    let filteredManga: IMangaCard[] = [];
-    if (query) {
-        const titleFilteredManga = manga.filter((m) => queryFilter(query, m));
-        const genreFilteredManga = manga.filter((m) => queryGenreFilter(query, m));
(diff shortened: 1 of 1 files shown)


src/components/library/LibraryMangaGrid.tsx
```diff
@@ -48,30 +48,25 @@ const queryGenreFilter = (query: NullAndUndefined<string>, { genre }: IMangaCard
...
-        }, {});
-        filteredManga = Object.values(unique);
-    }
-    filteredManga = (filteredManga.length ? filteredManga : manga).filter(
-        (m) => downloadedFilter(downloaded, m) && unreadFilter(unread, m),
-    );
-    return filteredManga;
-};
+): IMangaCard[] =>
+    performSearch(query, manga).filter((m) => downloadedFilter(downloaded, m) && unreadFilter(unread, m));
```
akabhirav, line 69: filtering the manga, before calling performSearch should increase the performance for large libraries

akabhirav, line 69:
```suggestion
    performSearch(query, manga.filter((m) => downloadedFilter(downloaded, m) && unreadFilter(unread, m)));
```
something like this

schroda: you're right it's not optimal currently.
I think the best thing should be to perform search and filter on each manga in one go.
I'll open a new pr for that

schroda: nervermind, changed it in this pr

akabhirav (approved):
LGTM

PR #252, opened 2023-02-27
Feature/extension card cleanup

schroda:
preparation for i18n usage + cleanup

### DIFF
--- a/src/components/ExtensionCard.tsx
+++ b/src/components/ExtensionCard.tsx
@@ -38,6 +38,18 @@ type InstalledStates = ExtensionAction | ExtensionState;
 
 const InstalledState = { ...ExtensionAction, ...ExtensionState } as const;
 
+const EXTENSION_ACTION_TO_STATE_MAP: { [action in ExtensionAction]: ExtensionState } = {
+    [ExtensionAction.UPDATE]: ExtensionState.UPDATING,
+    [ExtensionAction.UNINSTALL]: ExtensionState.UNINSTALLING,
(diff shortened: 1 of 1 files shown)


src/components/ExtensionCard.tsx
```diff
@@ -21,64 +21,74 @@ interface IProps {
...
-                uninstall();
-                setTimeout(() => window.location.reload(), 3000);
+            case ExtensionAction.INSTALL:
+            case ExtensionAction.UPDATE:
+            case ExtensionAction.UNINSTALL:
+                requestExtensionAction(installedState).catch(() => {});
                 break;
-            case 'uninstall':
-                uninstall();
+            case ExtensionState.OBSOLETE:
```
akabhirav, line 90: @schroda I wanted to know if this is the same mechanism used in the tachiyomi app. Last time this happend it was confusing to me. When I clicked on the obsolete button it started to uninstall the extension. Should we change this behaviour so that the user knows what they are doing by clicking on this?

schroda: no idea how that works in tachiyomi, never used that app

> Maybe just saying that the extension is obsolete as a label and then giving the uninstall button should work. I remember tachiyomi app doing that once, long time ago

another option could also be to add a group "Obsolote" (or some other name) in the "Source" screen and have the "action" be "Uninstall"

but this can/should be done in a separate pr

akabhirav: Agreed, this should be taken up in another PR, rest of the things look good

akabhirav (approved):

PR #253, opened 2023-03-04
Replaces Sort by ID with Date Added

akabhirav:
## Sort by Date Added
This sort helps you find the latest manga you've added to your Library

## Why replace Sort by ID
Sort by ID, sorted the manga with an internally generatec ID in backend. This is not at all helpful for users as this might seem very random to them and is random for intents and purposes for them. This sort can used in best case for debugging, but there are better ways for that.

## Updating the IManga type
This PR also updates the IManga type with all the latest data that it's getting from the backend. This can be very helpful for features that are planned for the future

## .nvmrc
This PR also adds ,nvmrc so it's easier for developers to switch to the relevant node version while developing and testing


schroda (approved):
LGTM

not really important but personally I would have added nvmrc with a different pr since it doesn't really have anything to do with the context of this one

akabhirav:
> not really important but personally I would have added nvmrc with a different pr since it doesn't really have anything to do with the context of this one

Yeah, sorry about that. I was getting lazy and included it here since it was a small file change

schroda:
no worries, if you're lazy, what am I then 😆, look at all your PR descriptions 👍

PR #255, opened 2023-03-05
Fix/update react scripts dependency

schroda:
upgrade to latest "react-script" version fixes the build error "error:0308010C:digital envelope routines::unsupported"

the latest version of "react-script" uses "eslint" v8, causing an error when building "... ERROR in Plugin "react" was conflicted between ".eslintrc.js" >> eslint-config-airbnb ...", thus, "eslint" has to be updated as well to be able for the latest "react-script" version to work without errors.


akabhirav (approved):
LGTM

PR #257, opened 2023-03-06
[Cleanup] Add enums and types

schroda:


schroda:
that is actually not the text that gets rendered, looks like I've missed it some hardcoded strings in `langauge.tsx` `ISOLanguages`

I'll fix/improve it in another pr

schroda:
added the changes to this pr

extracted the "default languages" (all, other, local source) from the "ISOLanguages" and also translating them using i18n

AriaMoradi:
We have to merge and lock translation strings before merging commits on the main repo and then pull on weblate

PR #261, opened 2023-03-14
[ESLint] Disable "no-shadow" rule

schroda:
Not sure what benefits this adds but every IDE (and dev should be) is able to tell these variables apart, and it's annoying having to find some different name

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,8 @@ module.exports = {
         'react/require-default-props': 'off',
         'react/function-component-definition': 'off',
 
+        '@typescript-eslint/no-shadow': 'off',
+
         'react/no-unstable-nested-components': [
             'error',
             {


AriaMoradi:
I think we should keep it on to explicitly avoid issues, there is always a better name, you just have to find it.

schroda:
mh, I'm not sure. Personally I feel like it's too restricting but on the other hand, for people that don't understand 100% what they're doing that can prevent issues ofc.

Here https://github.com/Suwayomi/Tachidesk-WebUI/pull/262/files#diff-737b8b1cc66819ce17acedfc417e454ecadd16af5a244118fbc0171928dcf3c9 (line 53) e.g. I feel like it's only annoying and doesn't really add anything.

Seems also like some others have also felt the same way since "/* eslint-disable @typescript-eslint/no-shadow */" can be found in the code base

AriaMoradi:
> Seems also like some others have also felt the same way since "/* eslint-disable @typescript-eslint/no-shadow */" can be found in the code base

3 files disable it, alright I'm convinced. Update them in this PR since it will be globally disabled.

martinek:
TBH, all of those cases could be solved simply. There is no reason for those conflicts to be there.

schroda:
The one from me was just copy&paste so there are only 2 and yes they can be fixed simply, I just don't like the rule 🤣

but yeah, I'll just close the pr

PR #296, opened 2023-05-14
Feature/update to swr version 2.1.5

schroda:


AriaMoradi (approved):
I'd name the merge commit something like "update to version 2.x" to signify that it's a major version bump and is handling some breaking changes.

PR #297, opened 2023-05-15
Feature/streamline backend requests

schroda:


AriaMoradi:
Good idea, this will help a lot to prepare for migrating to graphql.

PR #310, opened 2023-05-19
Fix/global search show request error

schroda:


schroda:
- [ ] merge after #308

PR #323, opened 2023-06-03
Add fit page to window reader setting

Alexandre-P-J:
## Description
The following change adds a new option to the reader settings called fit page to window. This feature allows users to display pages maximally fitted to the reader window without overflowing vertical or horizontaly.
The behavior is similar to the current page fitting of the two-page or horizontal scroll reader modes. The reader style where the change makes more sense is in both single page modes but it is also useful in webtoon and continues vertical modes.
### Single page
image
### Vertical
image

### Single page
image
### Vertical
image
### Single page
image
### Vertical
image

### DIFF
--- a/src/components/navbar/ReaderNavBar.tsx
+++ b/src/components/navbar/ReaderNavBar.tsx
@@ -261,6 +261,7 @@ export default function ReaderNavBar(props: IProps) {
                             showPageNumber={settings.showPageNumber}
                             loadNextOnEnding={settings.loadNextOnEnding}
                             skipDupChapters={settings.skipDupChapters}
+                            fitToWindow={settings.fitToWindow}
                             readerType={settings.readerType}
                         />
                     </Collapse>
--- a/src/components/navbar/ReaderNavBar.tsx
+++ b/src/components/navbar/ReaderNavBar.tsx
@@ -261,7 +261,7 @@ export default function ReaderNavBar(props: IProps) {
                             showPageNumber={settings.showPageNumber}
                             loadNextOnEnding={settings.loadNextOnEnding}
                             skipDupChapters={settings.skipDupChapters}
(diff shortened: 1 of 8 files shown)


src/components/navbar/ReaderNavBar.tsx
```diff
@@ -261,6 +261,7 @@ export default function ReaderNavBar(props: IProps) {
                             showPageNumber={settings.showPageNumber}
                             loadNextOnEnding={settings.loadNextOnEnding}
                             skipDupChapters={settings.skipDupChapters}
+                            fitToWindow={settings.fitToWindow}
```
schroda, line 264: I'd rename it to e.g. `fitPageToWindow`, to make it 100% clear

schroda (changes requested):
as you mentioned in discord, I'd say the setting should be hidden for reader modes where it doesn't do anything

src/components/navbar/ReaderNavBar.tsx
Alexandre-P-J: agree, fixed in next commit

Alexandre-P-J:
This last commit also hides the option in reader modes where it doesn't do anything.

src/i18n/locale/es.json
```diff
@@ -299,7 +299,8 @@
                 "static_navigation": "Navigación Estática",
                 "show_page_number": "Mostrar numero de pagina",
                 "reader_type": "Tipo de Lector",
-                "skip_dup_chapters": "Omitir capítulos duplicados"
+                "skip_dup_chapters": "Omitir capítulos duplicados",
+                "fit_to_window": "Ajustar página a la ventana"
```
schroda, line 303: nitpick, but you've forgot the translation key 😄

Alexandre-P-J: good catch

PR #417, opened 2023-09-05
Improvements on double page

rickymcmuffin:
This pull request changes how the double page works a bit. 

First, when there are two pages showing and the second page is the last page, it stops/goes to the next chapter.

Second, going to previous page has been improved as before, if you were on the last page and it was a double spread, it would make the last page single even if that wasn't correct.

This closes no issues

### DIFF
--- a/src/components/reader/pager/DoublePagedPager.tsx
+++ b/src/components/reader/pager/DoublePagedPager.tsx
@@ -24,7 +24,7 @@ const isSinglePage = (index: number, spreadPages: boolean[]): boolean => {
     // Page is single if it is last page
     if (index === spreadPages.length - 1) return true;
     // Page can not be single if it is not followed by spread
-    if (!spreadPages[index + 1]) return false;
+    // if (!spreadPages[index + 1]) return false;
     // Page is single if number of single pages since last spread is odd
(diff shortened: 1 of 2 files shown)


src/components/reader/pager/DoublePagedPager.tsx
```diff
@@ -100,7 +93,8 @@ export default function DoublePagedPager(props: IReaderProps) {
     }
 
     function nextPage() {
-        if (curPage < pages.length - 1) {
+        // Only go to next page if last page isn't displayed
+        if (curPage < pages.length - pagesDisplayed.current) {
```
schroda, line 97: this will cause the chapter not to get marked as read anymore, since for that, the last page has to be set as the current page

versionToServerVersionMapping.json
```diff
@@ -1,6 +1,6 @@
 [
   {
     "uiVersion": "PREVIEW",
-    "serverVersion": "r1353"
+    "serverVersion": "r1354"
```
schroda, line 4: this only has to be updated if you add changes that require a newer server version

schroda (changes requested):

versionToServerVersionMapping.json
rickymcmuffin: I changed it back to r1353

Issue #416, opened 2023-09-04
[Feature Request] Option to switch which page comes first in Double Page mode


rickymcmuffin:
## What feature should be added to Tachidesk?.
There should be a button or switch that says something like "Offset double spreads". When it is clicked it will change which page comes first. For example, if it is currently display page 9-10, It will instead display 10-11.

##  Why/Project's Benefit/Existing Problem
I personally enjoy reading double page as it gives the illusion of reading a manga through a book. However that illusion is broken when the pages aren't lined up to how they should be when reading it as a book. Adding this feature will make it easy to make sure the pages are lined up.

PR #418, opened 2023-09-07
Invert Double Page switch

rickymcmuffin:
I added a toggle that switches which page comes first in double page mode.

Before: 
beforeWebUI

After:
Screenshot 2023-09-07 112618

Here's with it off:
image

Here's with it on:
image

It is currently a work in progress. Features I want to add are:

- The toggle hides when not in double page mode.
- The pages switch instantly when clicking the toggle.

This closes the issue I made #416


rickymcmuffin:
Finished the changes. Let me know what you think!

schroda:
looks good, just a few things

since the setting only effects the double page reader, it should only be shown in case it is selected

I think "offset first page" would describes the behaviour better
the first time I read "invert double page" I didn't understand it and what it is supposed to do

not really relevant, but I would leave the "reader type" as the last setting, I think it looks better this way

rickymcmuffin:
I made all the changes.

schroda:
nice job improving the double page reader 👍

looks like there is also some issue with viewing a not downloaded/cached chapter with the double page reader, if you're interesed in taking a look at it

Issue #422, opened 2023-10-09
[Bug] Double Page doesn't show loading animation


rickymcmuffin:
## Device information
- Tachidesk version: v0.7.0-r1365
- Server Operating System: Windows 11 Pro
- Server Desktop Environment: N/A
- Server JVM version: java 17.0.7
- Client Operating System: Windows 11 Prp
- Client Web Browser: Firefox 118.0.1

## Steps to reproduce
1. Open a chapter of any manga in Double Page mode with a slow internet connection
2. Go to a page that hasn't been loaded yet

### Expected behavior
It shows a loading screen in place of the page that it is currently on.

### Actual behavior
It shows the last page that was loaded while displaying the new page number.

## Other details
When the first page is loaded but the second one isn't, it should show the first page with a loading screen in place of the second page.

PR #480, opened 2023-11-27
Adding page loading with Double Page Mode.

rickymcmuffin:
This will add a loading animation to each page in double page mode. What this fixes is the user gets feedback that the page has changed rather than it showing the previous page while it loads. This will also update the double page logic to only run after the pages have loaded.

This will fix issue #422 

This will also fix an issue that causes double page mode to only show one page if the first page is loaded but the second one isn't.

Loading screen when both pages are loading:
Screenshot 2023-11-27 094821

Loading screen when only the second page is loading:
Screenshot 2023-11-27 094931


schroda:
I needed the changes for #550

PR #562, opened 2024-01-17
Support configuring automatic downloads by category

chancez:
This PR implements changes to the UI to support configuring categories for automatic downloads https://github.com/Suwayomi/Suwayomi-Server/pull/832.

Adds a new "categories" section to the automatic downloads configuration.

Screenshot 2024-01-17 at 11 05 31 AM


src/components/settings/CategoriesSetting.tsx
```diff
@@ -19,35 +19,35 @@ import Button from '@mui/material/Button';
...
 import { ThreeStateCheckboxInput } from '@/components/atoms/ThreeStateCheckboxInput.tsx';
 import { makeToast } from '@/components/util/Toast.tsx';
-import { IncludeInUpdate } from '@/lib/graphql/generated/graphql.ts';
+import { IncludeOrExclude } from '@/lib/graphql/generated/graphql.ts';
 import { TCategory } from '@/typings.ts';
 import { requestManager } from '@/lib/requests/RequestManager.ts';
 import { CheckboxContainer } from '@/components/settings/globalUpdate/CheckboxContainer.ts';
 
-const booleanToIncludeInStatus = (status: boolean | null | undefined): IncludeInUpdate => {
+const booleanToIncludeInStatus = (status: boolean | null | undefined): IncludeOrExclude => {
```
schroda, line 27: the functions should probably be renamed as well (e.g. `booleanToIncludeOrExcludeStatus`)

src/components/settings/CategoriesSetting.tsx
```diff
@@ -78,9 +78,12 @@ const getCategoryUpdateInfo = (
     return categories.map((category) => category.name).join(', ');
 };
 
-export const GlobalUpdateSettingsCategories = () => {
+type CategoryIncludeField = keyof Pick<TCategory, 'includeInUpdate' | 'includeInDownload'>
+
+export const CategoriesSetting = (props: {includeField: CategoryIncludeField}) => {
```
schroda, line 83: I think `CategoriesIncludeExcludeSetting`, `CategoriesInclusionSetting`or something like that would be more descriptive

src/lib/graphql/generated/apollo-helpers.ts
schroda, line 1: looks like you've run the wrong script
you have to run `gql:codegen` which includes applying some changes to the created files after the actual codegen script has run

src/components/settings/CategoriesSetting.tsx
schroda, line 1: for the `includeInDownload` setting dialog the dialog text (`library.settings.global_update.categories.label.info`) should not get shown since it's incorrect

src/components/settings/CategoriesSetting.tsx
chancez: Yeah, I kinda just picked a name and figured it would come up during review. Naming is hard 😓

src/lib/graphql/generated/apollo-helpers.ts
chancez: Ah that explains why random things unrelated changed.

src/components/settings/CategoriesSetting.tsx
chancez: Good catch. I'm not super familiar with React, what's the right way to do that, and should that be controlled by the `includeField`, or through a separate prop?

Is something like this reasonable?
```
                    {
                      includeField == 'includeInUpdate' &&
                      <DialogContentText sx={{ paddingBottom: '10px' }}>
                          {t('library.settings.global_update.categories.label.info')}
                      </DialogContentText>
                    }
```

schroda: I'd just pass the dialog text as a prop and render it conditionally

```
{ 
  dialogText && (
    <DialogContentText sx={{ paddingBottom: '10px' }}>
      {dialogText}
    </DialogContentText>
  )
}
```

src/lib/graphql/generated/apollo-helpers.ts
schroda: mh, could be that something got added to the file which broke the script (`codegenFormatter.ts`)

src/components/settings/CategoriesSetting.tsx
chancez: Oh yeah that makes way more sense.

src/lib/graphql/generated/apollo-helpers.ts
chancez: Oh looks like `yarn gql:codegen-formatter` is what I needed to run.

src/components/settings/globalUpdate/GlobalUpdateSettings.tsx
```diff
@@ -47,7 +48,10 @@ export const GlobalUpdateSettings = () => {
         >
             <GlobalUpdateSettingsInterval />
             <GlobalUpdateSettingsEntries />
-            <GlobalUpdateSettingsCategories />
+            <CategoriesInclusionSetting
+              includeField='includeInUpdate'
+              dialogText={t('library.settings.global_update.categories.label.info')}
```
schroda, line 53: we should add a similar text for the download case

chancez: Updated. I also moved the other category text used in the CategoriesInclusionSetting component into a new top level `categories` section in the translation JSON files.

src/i18n/locale/en.json
```diff
@@ -27,6 +27,12 @@
       "set_categories": "Set categories"
     }
   },
+  "categories": {
```
schroda, line 30: I would add this under `"category"` as `"settings"`

```
"category": {
  ...
  "settings": {
    "inclusion": {
      "label": {
        "exclude": "Exclude: {{excludedCategoriesText}}",
        "include": "Include: {{includedCategoriesText}}"
      }
    }
}
```

src/i18n/locale/es.json
schroda, line 1: we only want to update the `en.json` language file, since this is the default one

chancez: But the structure changed, so these translations will be all busted if we don't move the settings, right?

src/i18n/locale/en.json
chancez: Done

src/i18n/locale/es.json
schroda: I think it will, but otherwise, open weblate changes have to always be commited to prevent merge conflicts, which will be forgotten at one point anyway resulting in issues

package-lock.json
schroda, line 1: the file has to be removed

schroda:
the `versionToServerVersionMapping.json` file has to be updated since these changes require at least server version `r1462`

package-lock.json
chancez: Oops. Fixed.

schroda (approved):

chancez:
Oops gotta fix the linter issues

PR #565, opened 2024-01-20
Make reader width configurable

chancez:
Adds a new setting  to configure the reader width. 

One unrelated change is I changed the adornment for NumberSetting from `startAdornment` to `endAdornment`. Having the unit (`%` ) in the front looked weird. I checked this effected other settings using the component and IMO it looked better as an endAdornment in all cases, but glad to make this a configurable property if that's preferred.

Screenshot 2024-01-20 at 11 18 24 AM

Screenshot 2024-01-20 at 11 17 14 AM

Screenshot 2024-01-20 at 11 17 00 AM

### DIFF
--- a/src/components/reader/ReaderSettingsOptions.tsx
+++ b/src/components/reader/ReaderSettingsOptions.tsx
@@ -10,10 +10,12 @@ import { List, ListItem, ListItemText, Switch } from '@mui/material';
 import Select from '@mui/material/Select';
 import MenuItem from '@mui/material/MenuItem';
 import { useTranslation } from 'react-i18next';
-import { IReaderSettings } from '@/typings';
+import { AllowedMetadataValueTypes, IReaderSettings } from '@/typings';
+import { NumberSetting } from '@/components/settings/NumberSetting.tsx';
+import { isFillsPageReaderType } from '@/components/reader/Page.tsx';
 
 interface IProps extends IReaderSettings {
-    setSettingValue: (key: keyof IReaderSettings, value: string | boolean) => void;
+    setSettingValue: (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => void;
 }
 
 export function ReaderSettingsOptions({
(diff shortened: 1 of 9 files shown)


schroda:
the setting does nothing in case "Fit page to window" is enabled or the "double page" or "vertical reader" is selected, so it would be best to hide it in these cases

chancez:
> the setting does nothing in case "Fit page to window" is enabled or the "double page" or "vertical reader" is selected, so it would be best to hide it in these cases

Is that the same as the conditions for `fitPageToWindowEligible`?

schroda:
yeah, something like that

chancez:
@schroda Ok, did a few updates, let me know what you think.

src/components/settings/NumberSetting.tsx
```diff
@@ -38,6 +39,7 @@ type BaseProps = {
     handleUpdate: (value: number) => void;
     showSlider?: never;
     disabled?: boolean;
+    sx?: SxProps<Theme>;
```
schroda, line 42: the name should indicate that it's only for the ListItemText

src/components/reader/ReaderSettingsOptions.tsx
```diff
@@ -87,7 +84,22 @@ export function ReaderSettingsOptions({
...
+                  dialogTitle={t('reader.settings.label.reader_width')}
+                  settingValue={`${readerWidth}%`}
+                  value={readerWidth}
+                  minValue={10}
+                  maxValue={100}
+                  defaultValue={100}
+                  valueUnit="%"
+                  showSlider
+                  handleUpdate={(width: number) => setSettingValue('readerWidth', width)}
+                  sx={{ display: 'flex', alignContent: 'center', justifyContent: 'space-between' }}
```
schroda, line 100: `alignContent` is incrorect, my bad, it has to be `alignItems`

chancez: Updated

package.json
```diff
@@ -35,6 +35,7 @@
     "@fontsource/roboto": "^5.0.8",
     "@mui/icons-material": "^5.15.2",
     "@mui/material": "^5.15.2",
+    "@mui/system": "^5.15.5",
```
schroda, line 38: is this needed? the `SxProps` can also be imported from `@mui/material`

chancez: I see, I can adjust the import

schroda (approved):

chancez:
Fixed CI

PR #567, opened 2024-01-24
Fix reader width

chancez:
The margin: auto on the Box container for the image was preventing 100% width to actually mean 100%.

Now that 100% is actually possible, I think fitPageToWindow makes more sense as a default. Additionally, since the image can fill 100% of the page, it can cover the ReaderNavBar, so set the z-index of the ReaderNavBar so it's rendered on top of the image and clickable.

## Before.

Notice 100%, is not actually taking 100% of the page width.

Screenshot 2024-01-23 at 7 07 17 PM

## After:

fit to page width
Screenshot 2024-01-23 at 7 04 21 PM

21% width
Screenshot 2024-01-23 at 7 04 14 PM

100% width
Screenshot 2024-01-23 at 7 04 06 PM

### DIFF
--- a/src/components/reader/DoublePage.tsx
+++ b/src/components/reader/DoublePage.tsx
@@ -33,7 +33,7 @@ export const DoublePage = forwardRef((props: IProps, ref: any) => {
     const spinnerStyle: SxProps<Theme> = {
         ...imgStyle,
         height: '100vh',
-        width: '70vw',
+        width: '50%',
         backgroundColor: '#525252',
     };
 
@@ -61,6 +61,7 @@ export const DoublePage = forwardRef((props: IProps, ref: any) => {
                 imgRef={imgRef}
                 spinnerStyle={{
                     ...spinnerStyle,
+                    width: 'calc(50% - 5px)',
                     marginLeft: '5px',
                 }}
                 imgStyle={imgStyle}
--- a/src/components/reader/DoublePage.tsx
+++ b/src/components/reader/DoublePage.tsx
@@ -39,7 +39,7 @@ export const DoublePage = forwardRef((props: IProps, ref: any) => {
             sx={{
                 display: 'flex',
                 justifyContent: 'center',
-                flexGrow: 1,
+                width: '100%',
             }}
         >
             <SpinnerImage
--- a/src/components/reader/DoublePage.tsx
+++ b/src/components/reader/DoublePage.tsx
@@ -6,26 +6,11 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-import { CSSProperties, forwardRef, useRef } from 'react';
+import { forwardRef, useRef } from 'react';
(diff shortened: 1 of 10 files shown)


schroda:
the single page does not take 100% of the width and is not centered when the height is < 100vh

chancez:
> the single page does not take 100% of the width and is not centered when the height is < 100vh

Ah your right, I gotta update the other pagers.

chancez:
So I fixed it for the single page layout. I was gonna update the DoublePage layouts to support adjustable width, but I cannot figure out how to make it take 100% of the width with horizontal layout because the flex direction has to be `row`.

chancez:
Ah I think I figured that out, it's related to the height set in the imageStyle in `DoublePage.tsx`

chancez:
Ok got double paged readers to work with the reader width setting too.

src/components/reader/DoublePage.tsx
```diff
@@ -68,10 +56,12 @@ export const DoublePage = forwardRef((props: IProps, ref: any) => {
                 imgRef={imgRef}
                 spinnerStyle={{
                     ...spinnerStyle,
-                    width: 'calc((100vw - 300px - 5px) * 0.5)',
                     marginLeft: '5px',
                 }}
-                imgStyle={imgStyles}
+                imgStyle={{
+                    ...imgStyle,
+                    marginLeft: '5px',
```
schroda, line 63: the pages are not supposed to have a gap

src/components/reader/Page.tsx
```diff
@@ -34,9 +34,6 @@ function imageStyle(settings: IReaderSettings): any {
     }, []);
     if (settings.fitPageToWindow || isFillsPageReaderType(settings.readerType)) {
         return {
-            display: 'block',
-            marginLeft: '7px',
```
schroda, line 38: this is removing the gap of the pages from the horizontal pager

src/components/reader/DoublePage.tsx
chancez: Oh, so you're right. I think I thought a gap was needed because of other changes I made that added one.

src/components/reader/Page.tsx
chancez: That explains why I thought we needed the gap in the double page

src/components/reader/DoublePage.tsx
schroda: the horizontal pager is fixed but gap in the double pager is still there

chancez: Ok, I think that should be fixed now.

schroda:
now the setting has no efffect on the pages of the double pager anymore 😅

chancez:
CSS is hard. `flex-grow: 1` fixes that. I'm not sure I fully understand why though, but I verified it at least maintains the aspect ratio of the images correctly.

schroda (approved):

PR #581, opened 2024-01-27
Update reader width live when updating in reader view

chancez:
Basically, instead of updating the width of the image after clicking `ok`, the width is adjusted live, which makes it easier to figure out the right value for the manga being viewed. 

Ideally we could revert the changes with the `cancel` button, but to do that requires tracking the value prior to the updates. With how the reader is designed currently, everything is using `settings` directly and so it's tricky to separate the settings being used to draw the component and the settings that are saved. So instead I've disabled the cancel button when `liveUpdate` is true, since settings are updated immediately.

While implementing this, I realized we can't update the backend every time we re-render when `liveUpdate` is true, so I've updated the logic for persisting settings to support skipping the API call to save the settings in the backend, and only perform the API it when the user clicks ok on the number setting.

### DIFF
--- a/src/components/settings/NumberSetting.tsx
+++ b/src/components/settings/NumberSetting.tsx
@@ -13,7 +13,7 @@ import TextField from '@mui/material/TextField';
 import { InputAdornment, ListItemText, Stack, SxProps, Typography, Theme } from '@mui/material';
 import DialogActions from '@mui/material/DialogActions';
 import Button from '@mui/material/Button';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import ListItemButton from '@mui/material/ListItemButton';
 import * as React from 'react';
@@ -39,6 +39,7 @@ type BaseProps = {
     showSlider?: never;
     disabled?: boolean;
     listItemTextSx?: SxProps<Theme>;
+    handleLiveUpdate?: (value: number) => void;
 };
 
 type PropsWithSlider = Omit<BaseProps, 'defaultValue' | 'minValue' | 'maxValue' | 'showSlider'> &
@@ -62,45 +63,49 @@ export const NumberSetting = ({
     handleUpdate,
     showSlider,
     disabled = false,
+    handleLiveUpdate,
     listItemTextSx: sx,
 }: Props) => {
     const { t } = useTranslation();
 
     const [isDialogOpen, setIsDialogOpen] = useState(false);
     const [dialogValue, setDialogValue] = useState(value);
+    const [originalValue, setOriginalValue] = useState(value);
 
-    const closeDialog = useCallback(
-        (resetValue: boolean) => {
+    const updateValue = useCallback(
+        (newValue: number, persist: boolean) => {
+            setDialogValue(newValue);
+            const didValueChange = newValue !== originalValue;
+            // Call handleUpdate if the value changed and 'persist' is true,
+            // otherwise call handleLiveUpdate if it's defined.
+            if (persist && didValueChange) {
+                handleUpdate(newValue);
+            } else if (handleLiveUpdate) {
+                handleLiveUpdate(newValue);
+            }
+        },
+        [originalValue, setDialogValue, handleLiveUpdate, handleUpdate],
+    );
+
+    const cancel = useCallback(() => {
+        updateValue(originalValue, true);
+        setOriginalValue(originalValue);
+        setIsDialogOpen(false);
+    }, [originalValue, handleUpdate]);
+
+    const resetToDefault = useCallback(() => {
+        if (defaultValue !== undefined) {
+            updateValue(defaultValue, true);
+            setOriginalValue(defaultValue);
             setIsDialogOpen(false);
+        }
+    }, [defaultValue, handleUpdate]);
 
-            if (resetValue) {
-                setDialogValue(value);
-            }
-        },
-        [value],
-    );
-
-    const closeDialogWithReset = useCallback(() => closeDialog(true), [closeDialog]);
-
-    const updateSetting = useCallback(
-        (newValue: number, shouldCloseDialog: boolean = true) => {
-            if (shouldCloseDialog) {
-                closeDialog(false);
-            }
-
-            const didValueChange = value !== newValue;
-            if (!didValueChange) {
-                return;
-            }
-
-            handleUpdate(newValue);
-        },
-        [value, handleUpdate, closeDialog],
-    );
-
-    useEffect(() => {
-        setDialogValue(value);
-    }, [value]);
+    const submit = () => {
+        updateValue(dialogValue, true);
+        setOriginalValue(dialogValue);
+        setIsDialogOpen(false);
+    };
 
     return (
         <>
@@ -114,7 +119,7 @@ export const NumberSetting = ({
                 />
             </ListItemButton>
 
-            <Dialog open={isDialogOpen} onClose={closeDialogWithReset}>
+            <Dialog open={isDialogOpen} onClose={cancel}>
                 <DialogContent>
                     <DialogTitle sx={{ paddingLeft: 0 }}>{dialogTitle}</DialogTitle>
                     {(!!dialogDescription || !!dialogDisclaimer) && (
@@ -158,7 +163,10 @@ export const NumberSetting = ({
                         autoFocus
                         value={dialogValue}
                         type="number"
-                        onChange={(e) => setDialogValue(Number(e.target.value))}
+                        onChange={(e) => {
+                            const newValue = Number(e.target.value);
+                            updateValue(newValue, false);
+                        }}
                     />
                     {showSlider ? (
                         <Slider
@@ -168,31 +176,22 @@ export const NumberSetting = ({
                             step={stepSize}
                             min={minValue}
                             max={maxValue}
-                            onChange={(_, newValue) => setDialogValue(newValue as number)}
+                            onChange={(_, newValue) => {
+                                updateValue(newValue as number, false);
+                            }}
                         />
                     ) : null}
                 </DialogContent>
                 <DialogActions>
                     {defaultValue !== undefined ? (
-                        <Button
-                            onClick={() => {
-                                setDialogValue(defaultValue);
-                                updateSetting(defaultValue, false);
-                            }}
-                            color="primary"
-                        >
+                        <Button onClick={resetToDefault} color="primary">
                             {t('global.button.reset_to_default')}
                         </Button>
                     ) : null}
-                    <Button onClick={closeDialogWithReset} color="primary">
+                    <Button onClick={cancel} color="primary">
                         {t('global.button.cancel')}
                     </Button>
-                    <Button
-                        onClick={() => {
-                            updateSetting(dialogValue);
-                        }}
-                        color="primary"
-                    >
+                    <Button onClick={submit} color="primary">
                         {t('global.button.ok')}
                     </Button>
                 </DialogActions>
--- a/src/screens/Reader.tsx
+++ b/src/screens/Reader.tsx
@@ -257,11 +257,13 @@ export function Reader() {
             .response.catch();
     };
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
         setSettings({ ...settings, [key]: value });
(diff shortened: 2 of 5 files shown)


src/screens/Reader.tsx
```diff
@@ -235,11 +235,13 @@ export function Reader() {
             .response.catch();
     };
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: Boolean = true) => {
```
schroda, line 238:
```suggestion
    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
```

src/components/settings/NumberSetting.tsx
```diff
@@ -98,6 +98,17 @@ export const NumberSetting = ({
...
+            handleUpdate(newValue);
+            if (handleSubmit) {
+                handleSubmit(newValue);
+            }
+            closeDialog(false);
+        },
+        [value, handleUpdate, handleSubmit, closeDialog],
+    );
+
     useEffect(() => {
```
schroda, line 112: this causes to many state updates during the live update

tbh, don't remember why I added this (I think to update the dialog value in case there was an update to it while the dialog was open?), probably can be completely removed

src/components/settings/NumberSetting.tsx
```diff
@@ -177,19 +199,21 @@ export const NumberSetting = ({
...
                             }}
                             color="primary"
                         >
                             {t('global.button.reset_to_default')}
                         </Button>
                     ) : null}
-                    <Button onClick={closeDialogWithReset} color="primary">
-                        {t('global.button.cancel')}
-                    </Button>
+                    {!liveUpdate ? (
```
schroda, line 209: the `cancel` behaviour can be kept by adding a new state (e.g. `originalValue`) which only gets updated when the change gets persisted, when canceling, the setting gets reset to the `originalValue`

schroda:
not sure if that is fixable with the way the reader and reader settings are implemented, but it does feel kinda laggy

src/components/settings/NumberSetting.tsx
chancez: I tried something like that, but the problem is at least in the way settings are passed as props, the value passed in initially is updated too. 

I basically had this, but the problem was `originalValue` was being updated too. Is it because I'm not using `useState`?
```
     const closeDialog = useCallback(
        (resetValue: boolean) => {
            setIsDialogOpen(false);

            if (resetValue) {
                setDialogValue(originalValue);
            }
        },
        [originalValue],
    );
```

I'll give it another try with state to see what happens.

chancez: Ok I figured this out

src/components/settings/NumberSetting.tsx
```diff
@@ -36,9 +36,11 @@ type BaseProps = {
     dialogDisclaimer?: string;
     valueUnit: string;
     handleUpdate: (value: number) => void;
+    handleSubmit?: (value: number) => void;
```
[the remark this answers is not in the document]
chancez: Ok, I changed it slightly, only 1 optional prop, just the callback and I tried to adapt your suggestion. LMK if I did it right or not

chancez:
@schroda Okay I took another stab, added `cancel` back and switched `handleUpdate` to the previous behavior and added a new optional `handleLiveUpdate` callback prop to update the setting without persisting.

src/components/settings/NumberSetting.tsx
```diff
@@ -39,12 +39,15 @@ type BaseProps = {
     showSlider?: never;
     disabled?: boolean;
     listItemTextSx?: SxProps<Theme>;
+    handleLiveUpdate?: (value: number) => void;
```
schroda, line 42: looking at it again, do we even need this new callback, couldn't it just be an additional parameter (e.g. `persist`) of `handleUpdate` that is only true if the ok button got clicked?

src/components/settings/NumberSetting.tsx
```diff
@@ -62,41 +65,50 @@ export const NumberSetting = ({
...
+            if (handleLiveUpdate) {
+                handleLiveUpdate(newValue);
             }
         },
-        [value],
+        [handleLiveUpdate],
     );
 
-    const closeDialogWithReset = useCallback(() => closeDialog(true), [closeDialog]);
+    const cancel = useCallback(() => {
```
schroda, line 87: `cancel` and  `resetToDefault` are basically the same function, should be merged

src/screens/Reader.tsx
```diff
@@ -235,11 +235,13 @@ export function Reader() {
             .response.catch();
     };
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist?: boolean) => {
```
schroda, line 238:
```suggestion
    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
```

src/screens/settings/DefaultReaderSettings.tsx
```diff
@@ -39,10 +39,12 @@ export function DefaultReaderSettings() {
 
     useSetDefaultBackTo('settings');
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
-        requestUpdateServerMetadata(convertToGqlMeta(metadata)! ?? {}, [[key, value]]).catch(() =>
-            makeToast(t('reader.settings.error.label.failed_to_save_settings'), 'warning'),
-        );
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist?: boolean) => {
```
schroda, line 42:
```suggestion
    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
```

src/components/settings/NumberSetting.tsx
schroda: what about this?

src/components/settings/NumberSetting.tsx
chancez: That feels like a weird thing to add to the generic `NumberSetting` component. Persistence isn't really something this component should need to worry about, no?

schroda: maybe `persist` was a bad name, more like `okClicked`, but nevermind me, I think having 2 callbacks is easier to use

src/components/settings/NumberSetting.tsx
```diff
@@ -39,12 +39,15 @@ type BaseProps = {
...
     disabled?: boolean;
     listItemTextSx?: SxProps<Theme>;
+    handleLiveUpdate?: (value: number) => void;
 };
 
 type PropsWithSlider = Omit<BaseProps, 'defaultValue' | 'minValue' | 'maxValue' | 'showSlider'> &
     Required<Pick<BaseProps, 'defaultValue' | 'minValue' | 'maxValue'>> & { showSlider: true };
 
-type Props = BaseProps | PropsWithSlider;
+type LiveUpdateProps = (BaseProps | PropsWithSlider) & { handleLiveUpdate: (value: number) => void };
```
schroda, line 48: since this other prop was removed I don't think the extra type is need anymore

src/components/settings/NumberSetting.tsx
```diff
@@ -62,41 +65,50 @@ export const NumberSetting = ({
...
-            const didValueChange = value !== newValue;
-            if (!didValueChange) {
-                return;
-            }
-
-            handleUpdate(newValue);
-        },
-        [value, handleUpdate, closeDialog],
-    );
+    const submit = useCallback(() => {
```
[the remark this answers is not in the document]
chancez: So do I just use a regular callback function?

src/components/settings/NumberSetting.tsx
chancez: Tbh I don't really understand `useEffect`. So just remove it?

src/components/settings/NumberSetting.tsx
chancez: They're similar, but not the same. They call the same functions, but the values and conditions are different. Do you have any recommendations about they should be combined?

chancez:
Pushed after rebasing and fixing what I could so I can get a re-review, but need to test things again.

chancez:
Ok should be good, I tested and it seems to be working as expected.

src/components/settings/NumberSetting.tsx
schroda: yeah, if everything works, just remove it

useEffect gets called after every render
here it would reset the current tmp dialog value to the original value in case it has changed

src/components/settings/NumberSetting.tsx
schroda: you could do a function that gets passed an optional value to reset to as an argument

src/components/settings/NumberSetting.tsx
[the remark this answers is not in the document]
schroda: you can just remove useCallback

src/components/settings/NumberSetting.tsx
```diff
@@ -62,45 +63,50 @@ export const NumberSetting = ({
...
-                closeDialog(false);
-            }
+    const cancel = useCallback(() => {
+        setDialogValue(originalValue);
+        setOriginalValue(originalValue);
+        handleUpdate(originalValue);
+        setIsDialogOpen(false);
+    }, [originalValue, handleUpdate]);
 
-            const didValueChange = value !== newValue;
```
schroda, line 91: this should be kept to prevent unnecessary requests, currently clicking cancel sends a requests for the same value

src/components/settings/NumberSetting.tsx
```diff
@@ -62,45 +63,50 @@ export const NumberSetting = ({
...
+            handleUpdate(defaultValue);
+            setIsDialogOpen(false);
+        }
+    }, [defaultValue, handleUpdate]);
 
-    useEffect(() => {
-        setDialogValue(value);
-    }, [value]);
+    const submit = () => {
+        const didValueChange = dialogValue !== originalValue;
```
schroda, line 102: this should be moved to `handleUpdate` since `cancel` and `resetToDefault` value currently send a request for the same value

src/screens/Reader.tsx
```diff
@@ -257,11 +257,13 @@ export function Reader() {
             .response.catch();
     };
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
         setSettings({ ...settings, [key]: value });
-        requestUpdateMangaMetadata(manga, [[key, value]]).catch(() =>
-            makeToast(t('reader.settings.error.label.failed_to_save_settings'), 'warning'),
-        );
+        if (persist === undefined || persist) {
```
schroda, line 262:
```suggestion
        if (persist) {
```

src/screens/settings/DefaultReaderSettings.tsx
```diff
@@ -39,10 +39,12 @@ export function DefaultReaderSettings() {
 
     useSetDefaultBackTo('settings');
 
-    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes) => {
-        requestUpdateServerMetadata([[key, value]]).catch(() =>
-            makeToast(t('reader.settings.error.label.failed_to_save_settings'), 'warning'),
-        );
+    const setSettingValue = (key: keyof IReaderSettings, value: AllowedMetadataValueTypes, persist: boolean = true) => {
+        if (persist === undefined || persist) {
```
schroda, line 43:
```suggestion
        if (persist) {
```

src/components/settings/NumberSetting.tsx
chancez: Hmm, previously this was in the `updateSetting` function and moving this into the `handleUpdate` prop means that all existing uses of `NumberSetting` will need to be updated, is that what's wanted?

schroda: you're right, it should be handled in this component and not the prop, I mistook this for `updateValue`

I think it would be best to e.g. add a `persist` parameter to `updateValue` and then call `handleUpdate` instead of `handleLiveUpdate` in case `true` gets passed
with this everything else has to call `updateValue` to be able to change it

then the `didValueChange` check could be move there to prevent unnecessary requests/updates for all value updates and not just from `submit`

schroda (approved):

PR #694, opened 2024-04-01
Fix tracker refresh mutation

taos15:
This pull request fixes the tracker refresh mutation by adding the necessary code to fetch the track record.


taos15 (commented):
I just realized that [PR 692](https://github.com/Suwayomi/Suwayomi-WebUI/pull/692/files#diff-9650bd6ad5745b4d2224c5dfc1c6029e1612def87c2d0fdbd006fe2352ffdd04) created the `TRACKER_FETCH_BIND` I will use that mutation instead, I remove the mutation from the useEffect, since it was causing multiple mutations when the page re-rendered.

taos15:
I will use the just implemented TRACKER_FETCH_BIND mutation

PR #708, opened 2024-04-05
Feature/mark-chapters-as-read-based-on-tracker-history

taos15:
This pull request includes the following changes:

- Refactor: creates utility functions to work with lists

- Refactor(chapters): removes commented out code and implements utility function to clean up the file

- Refactor TrackMangaButton to update local source if the tracker last read is higher than local

- Merge branch 'master' of github.com:Suwayomi/Suwayomi-WebUI into feature/mark-chapters-as-read-based-on-tracker-history

- Refactor(packe.json): adds --host flag to the vite command to be able to run the dev server remotely


taos15:
This PR implement [feature request 699](https://github.com/Suwayomi/Suwayomi-WebUI/issues/699)

taos15 (commented):
I need to add logic to support partial chapters. like 56.1

src/components/chapter/ChapterActionMenuItems.tsx
```diff
@@ -118,15 +119,7 @@ export const ChapterActionMenuItems = ({
...
-
-            const index = allChapters.findIndex(({ id: chapterId }) => chapterId === chapter.id);
-
-            const isFirstChapter = index + 1 > allChapters.length - 1;
-            if (isFirstChapter) {
-                return [];
-            }
-
-            return allChapters.slice(index + 1);
+            return getPreviousChapters(chapter.id, allChapters);
```
schroda, line 122: use the `Chapters` class - same for all the added util functions

schroda (commented):
this should be an action in the options menu of an active track binding (`TrackerActiveCard`)
it also should not fetch the chapter list and only sync the known chapters

taos15:
Since it check which tracker has the more chapter read, should the action be in the modal instead of the trackerCard? Like a button somewhere in the model (sync read)? 

For the chapters, you mean work with the current chapter that the manga has and not try to pull the latest? I might have done without knowing, but I did not know I was fetching the chapters. My intent was to only fetch the trackers history (trackerBind), and use the current chapters list to comprare (I use the requestManager because I though it return it from cache, kid of a cache store).

src/components/manga/TrackMangaButton.tsx
```diff
@@ -28,8 +29,67 @@ export const TrackMangaButton = ({ manga }: { manga: TManga }) => {
 
     const loggedInTrackers = Trackers.getLoggedIn(trackerList.data?.trackers.nodes ?? []);
     const trackersInUse = Trackers.getLoggedIn(Trackers.getTrackers(mangaTrackers));
+    const mangaChaptersQuery = requestManager.useGetMangaChapters(manga.id, {});
```
taos15, line 32: I this the line you are talking about that is fetching the chapters? @schroda

schroda: no, it's this

```
        // Fetch new chapters if behind tracker
        if (localBehindTracker) {
            await requestManager.getMangaChaptersFetch(manga.id, { awaitRefetchQueries: true }).response;
        }
```

taos15: Ok. If the local manga is behind (last chapter is 100 but last read in tracker is 103) there wont be a way to find the last chapter that match the tracker. In that case, do you propose to not get the last chapter and instead mark every as read?. Since in the example, the last local chapter is 100 and the last tracker read is 103, every before 103 should be read, in this case all the local chapters should be marked as read?

schroda: yes, I am

taos15: Ok. I can do that.

PR #721, opened 2024-04-09
feat(request class): adds header to send cookies, and other auths with every request

taos15:
This pull request adds a header to send cookies and other authentication information with every request. 
- It updates to the GraphQLClient and RestClient classes to handle credentials and include them in the requests.

**References**
- The `Access-Control-Allow-Origin` needs to be set, without using wildcare (`*`), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin).
- [Apollo Client Authentication](https://www.apollographql.com/docs/react/networking/authentication/) documentation.
- [Traefik cors](https://doc.traefik.io/traefik/middlewares/http/headers/#cors-headers) example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ build/*
 tools/scripts/github_token.json
 
 src/lib/graphql/schema.json
+# local dev
+compose.*
+pnpm*.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,3 @@ build/*
 tools/scripts/github_token.json
 
 src/lib/graphql/schema.json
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -9,3 +9,6 @@ build/*
 tools/scripts/github_token.json
 
 src/lib/graphql/schema.json
+# local dev
```
schroda, line 12: if you add stuff that only exists for you, then add this to your local gitignore

taos15: removed.

PR #732, opened 2024-04-14
fix(manifest.json): fetch the manifest.json and cache when the app first load to eliminate 401 when using a reverse proxy like nginx

taos15:
This pull request fixes the issue where the manifest.json was not being fetched and cached when the app first loaded, causing a 401 error when using a reverse proxy like nginx. The `fetchAndCacheManifest` function has been added to the `App` component's `useEffect` hook to fetch and cache the manifest.json.


taos15 (commented):
Implemented in [PR 737](https://github.com/Suwayomi/Suwayomi-WebUI/pull/737)

PR #736, opened 2024-04-14
feat(app.tsx): refactors the app.tsx routes to lazy load the components.

taos15:
This pull request refactors the app routes to lazy load the components, improving performance and reducing initial bundle size.


taos15 (commented):
This is a WIP, it is open for people to colaborate. Adding code split will be a lot of work. We will need to check every screen/component to see what can be lazy load. Since the not-default-exort eslint rule was use in the app, every component that is lazy load needs to be refactor to default exports (there are ways to lazy load name export, but it is more work).

schroda:
default exports won't be introduced to the code base again, either do it with named exports or don't do it at all
and don't create a massive pr which introduces lazy loading everywhere, split it up and create small prs

PR #793, opened 2024-09-03
Fix overflow on Continuous Horizontal view

BrutuZ:
Fixes a vertical overflow on the Continuous Horizontal View of the reader. Especially noticeable when the default font size is overridden on the browser (I use 20 instead of 16)

| Before | After |
| --- | --- |
| image | image |
| Vertical Overflow | No Vertical Overflow |

### DIFF
--- a/src/screens/Reader.tsx
+++ b/src/screens/Reader.tsx
@@ -553,7 +553,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: '100vh',
+                minHeight: 'calc(100vh - 1rem)',
                 marginLeft: `${navBarWidth}px`,
             }}
         >
--- a/src/screens/Reader.tsx
+++ b/src/screens/Reader.tsx
@@ -42,6 +42,7 @@ import { GET_CHAPTERS_READER } from '@/lib/graphql/queries/ChapterQuery.ts';
 import { GET_MANGA_READER } from '@/lib/graphql/queries/MangaQuery.ts';
 import { TMangaReader } from '@/lib/data/Mangas.ts';
 import { CHAPTER_READER_FIELDS } from '@/lib/graphql/fragments/ChapterFragments.ts';
+import { useResizeObserver } from '@/util/useResizeObserver';
 
 type TChapter = GetChaptersReaderQuery['chapters']['nodes'][number];
 
@@ -487,6 +488,12 @@ export function Reader() {
         openNextChapter(ChapterOffset.PREV);
     }, [openNextChapter]);
 
+    const [scrollbarHeight, setScrollbarHeight] = useState(0);
+    useResizeObserver(
+        document.documentElement,
+        useCallback(() => setScrollbarHeight(window.innerHeight - document.documentElement.clientHeight), []),
+    );
+
     if (isLoading) {
         return (
             <Box
@@ -553,7 +560,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: 'calc(100vh - (100vh - 100%))',
+                minHeight: `calc((100vh - ${scrollbarHeight}px)`,
                 marginLeft: `${navBarWidth}px`,
             }}
         >
--- a/src/screens/Reader.tsx
+++ b/src/screens/Reader.tsx
@@ -553,7 +553,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: `calc(100vh - (100vh - 100%))`,
+                minHeight: 'calc(100vh - (100vh - 100%))',
                 marginLeft: `${navBarWidth}px`,
             }}
         >
--- a/src/screens/Reader.tsx
+++ b/src/screens/Reader.tsx
@@ -560,7 +560,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: `calc((100vh - ${scrollbarHeight}px)`,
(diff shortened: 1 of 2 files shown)


schroda:
the problem is the horizontal scrollbar height which the `100vh` does not include, making `1rem` not the ideal size to subtract since it only works as long as the actual scrollbar height is less than 16px and additionally, the scrollbar height might be less than 16px which would make the page take up less than 100% of the available height

instead, the scrollbar height should be set to the same fixed value as the width (`theme.tsx`), which is currently not done, and then the `SCROLLBAR_WIDTH` (probably should also be renamed, e.g. `SCROLLBAR_SIZE`) can be used for the calculation

browsers where the scrollbars do not affect the available height also need be considered, since there would be empty space due to pages not taking the full height because the non-existing scrollbar height gets subtracted
so it has to be considered if a scrollbar exists which reduces the available height and only then the scrollbar height should be subtracted

BrutuZ:
I tried getting the scrollbar size on `theme.ts` but the snippet returns falsy (I'm guessing there's nothing rendered at that point of the stack), so I did it locally instead

schroda:
what I meant was the `export const SCROLLBAR_WIDTH = 14;` which is in `theme.ts` where the toolbar styling is defined

```ts
MuiCssBaseline: {
                    styleOverrides: `
                        *::-webkit-scrollbar {
                          width: ${SCROLLBAR_WIDTH}px;
                          # only width is set and height is missing but should be added so that both the vertical and horizontal scrollbar have the same size
                          # height: ${SCROLLBAR_WIDTH}px;
                        }
                        ...
```

regarding the scrollbar detection 
```ts
const scrollbarHeight = window.innerHeight - document.documentElement.clientHeight;
```

this can fail since there is no gurantee that the scrollbar is already visible by the time the component renders, so the better option would be to listen to size changes of e.g. the `document.documentElement` (`useResizeObserver` can be used for that) and to update some state (e.g. `scrollbarSize` or `isScrollbarVisible`)

src/screens/Reader.tsx
```diff
@@ -553,7 +553,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: '100vh',
+                minHeight: `calc(100vh - (100vh - 100%))`,
```
schroda, line 556: does this work?

BrutuZ: It does, in the sense that `100vh` nets me that smidge of overflow and this doesn't.

BrutuZ:
> what I meant was the export const SCROLLBAR_WIDTH = 14; which is in theme.ts where the toolbar styling is defined

That's what I tried to change for the script snippet.
Worth noting that Firefox ignores that scrollbar styling.

> regarding the scrollbar detection
> 
> ```ts
> const scrollbarHeight = window.innerHeight - document.documentElement.clientHeight;
> ```
> this can fail since there is no guarantee that the scrollbar is already visible by the time the component renders, so the better option would be to listen to size changes of e.g. the `document.documentElement` (`useResizeObserver` can be used for that) and to update some state (e.g. `scrollbarSize` or `isScrollbarVisible`)

Well, now things went way out of my league, I'm not familiar with Vue or to understand that 😅

src/screens/Reader.tsx
schroda: I think thats because this results in `calc(100vh - (100vh - 0))` since the parent element does not have a height set, so it takes 100% of 0 - which at the end makes this a min height of 0

schroda:
> Worth noting that Firefox ignores that scrollbar styling.

yeah, firefox is also one of the browsers were the scrollbar does not have any affect on the available width/height

> Well, now things went way out of my league, I'm not familiar with Vue or to understand that 😅

basically, by the time the Page renders `scrollbarHeight  (const scrollbarHeight = window.innerHeight - document.documentElement.clientHeight;)` can still be 0, because the element which causes the scrollbar has not been rendered yet

the reason this still works most of the time, is because the page gets rerendered so often that at some point in time the scrollbar is visible, therefore `scrollbarHeight` would not be 0 anymore
however, in case the page never gets rerendered while the scrollbar is already visible, the scrollbar size would never be subtracted from the available height

src/screens/Reader.tsx
BrutuZ: `getComputedStyle()` returned `100%` for it on the browser's Developer Tools
image

BrutuZ:
Thanks, I got so far as to understand the theory of why that's flawed (typical race condition), it's the part about attaching an observer and updating states that lost me.

src/screens/Reader.tsx
schroda: interesting, nonetheless, 100% of 0 height of the parent is still 0

schroda:
you can just reuse an existing hook `useResizeObserver`, it would look like this
```ts
...
const [scrollbarHeight, setScrollbarHeight] = useState(0)
useResizeObserver(
  # elment to observer size changes
  document.documentElement,
  # handle size change by calculating the scrollbar height and updating the components scrollbarHeight state which triggers a rerender
  useCallback(() => setScrollbarHeight(window.innerHeight - document.documentElement.clientHeight),
  []
)
...

```

src/screens/Reader.tsx
BrutuZ: I tried changing it to both `max-height` and `height` just to see if it would "disappear" with 0 height, it didn't. Computed style for `height` was even in absolute units
image

schroda: you can see it by adding `max-height` with both % and vh values, the % based value does not do anything while the vh one does shrink the pages in size

BrutuZ:
Something like that https://github.com/Suwayomi/Suwayomi-WebUI/pull/793/commits/1189852052f023f2e9ad4799fd7cb9b3f57ac874 ?

src/screens/Reader.tsx
schroda: this needs to be changed to the same fix as in `Page` (`calc(100vh - ${scrollbarHeight}px)`)

as it is now it breaks the single and double page mode while `fit to page` is disabled and the `reader width` is below 100%, in that case the page is not vertically centered anymore

src/screens/Reader.tsx
```diff
@@ -553,7 +560,7 @@ export function Reader() {
                 alignItems: 'center',
                 justifyContent: 'center',
                 minWidth: `calc((100vw - (100vw - 100%)) - ${navBarWidth}px)`, // 100vw = width excluding scrollbar; 100% = width including scrollbar
-                minHeight: '100vh',
+                minHeight: `calc((100vh - ${scrollbarHeight}px)`,
```
schroda, line 563:
```suggestion
                minHeight: `calc(100vh - ${scrollbarHeight}px)`,
```

src/screens/Reader.tsx
schroda: looks good

PR #816, opened 2024-11-11
feat(spinnerImage): make image load more snappy, remove blank state

castrix:
I noticed that there are blank state (couple miliseconds rendering) when changing pages (especially when using local source), this commit is to make the page navigation more snappy and seamless

before
before

after
after


schroda:
all images are loaded through a queue to be able to prevent the server from getting spammed with image requests by aborting pending requests

this is necessary since, unless images have been cached/downloaded by the server, they will be requested from the sources, which are rate limited causing the server to potentially get unresponsive depending on the amount of requests

the blinking shown in the before gif should also have been already fixed in the preview version

PR #885, opened 2025-02-25
Support non-Latin characters in library search function

crystailx:
### What does this PR do?

This PR updates the enhancedCleanup function to properly handle non-Latin characters (e.g., Japanese, Chinese) by:
- Using a Unicode regex (\p{L}\p{N}\s) to ensure multilingual support while still removing unnecessary symbols/punctuations to keep text clean for search.
- Applying Unicode NFKC normalization to unify full-width and half-width character variations.

### Why is this needed?

The previous implementation (replace(/[^a-zA-Z0-9\\s]+/g, ' ')) removed all non-Latin characters, preventing proper search functionality for titles written in languages other than English.
Now, users can search for titles more flexibly, even if they contain diacritics, full-width characters, or non-Latin scripts.

### Example
| Input | Old Output | New Output |
| ----- | ----------- | ------------ |
| "ＡＢＣ123" | "123" | "abc123" |
| "製品-123!" | "123" | "製品 123" |
| "Café Latte" | "caf latte" | "café latte" |
| "カフェ・ラテ" | "" | "カフェ ラテ" |

### DIFF
--- a/src/util/Strings.ts
+++ b/src/util/Strings.ts
@@ -10,7 +10,8 @@ export const baseCleanup = (str: string) => str.toLowerCase().trim();
 
 export const enhancedCleanup = (str: string): string =>
     baseCleanup(str)
-        .replace(/[^a-zA-Z0-9\\s]+/g, ' ')
+        .normalize('NFKC')
+        .replace(/[^\p{L}\p{N}\s]+/gu, ' ')
         .trim();
 
 export const reverseString = (str: string, separator: string = ''): string =>
--- a/src/util/Strings.ts
+++ b/src/util/Strings.ts
@@ -11,7 +11,7 @@ export const baseCleanup = (str: string) => str.toLowerCase().trim();
 export const enhancedCleanup = (str: string): string =>
     baseCleanup(str)
         .normalize('NFKC')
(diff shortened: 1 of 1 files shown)


src/util/Strings.ts
```diff
@@ -10,7 +10,8 @@ export const baseCleanup = (str: string) => str.toLowerCase().trim();
 
 export const enhancedCleanup = (str: string): string =>
     baseCleanup(str)
-        .replace(/[^a-zA-Z0-9\\s]+/g, ' ')
+        .normalize('NFKC')
+        .replace(/[^\p{L}\p{N}\s]+/gu, ' ')
```
schroda, line 14:
```suggestion
        .replace(/[^\p{L}\p{N}]+/gu, ' ')
```
the old regex unnecessarily contained `\\s` which did nothing, but `\s` leads to multiple spaces in case a char gets replaced

e.g.
`Test - 123`
old: `Test 123`
new: `Test   123`

crystailx: Nice catch! You're right, I didn't notice that.
I've updated the regex as suggested. Thanks for pointing it out!  🙌

Issue #902, opened 2025-03-17
[Bug] Webtoon-mode of new player jumps unpredictably in certain situations


cpiber:
## Device information
- Suwayomi-WebUI version: r2401
- Suwayomi-Server version: v1.1.1-r1701
- Server Operating System: FreeBSD 14.2-RELEASE-p2
- Server Desktop Environment: N/A
- Server JVM version: openjdk version "21.0.6" 2025-01-21
- Client Operating System: Ubuntu 24.04.1
- Client Web Browser: Firefox 136.0.1

## Info
Both Webtoon and Continous Scroll are implemented via infinite scroll. When reading many chapters at once, this does not unload previous chapters, which can get really weird behavior (very long pages). This is especially problematic when the reader mode is changed or the window is resized - since pages are not rendered properly, it often happens that the scroll position is misinterpreted and the reader changes chapter wildly in such situations (I've also had happen that the reader fails to load any chapter, just showing the transition, and I have to use the selector to find the chapter I was at).

The same thing can also happen if the users scrolls too far while the images are still loading (especially with tall pages as is common in web toons).

 _Originally posted by @cpiber in [#778](https://github.com/Suwayomi/Suwayomi-WebUI/issues/778#issuecomment-2730440214)_

## Steps to reproduce
1. Open a Manga in Webtoon mode (should have a good number of pages, or more chapters)
2. Scroll down until the scroll bar becomes reasonably small
3. Make browser window smaller or change reader mode

### Expected behavior
The reader to adjust and stay on the selected page

### Actual behavior
Reader position jumps elsewhere proportional to resize amount

### Additonal details
It seems the reader does not react to resize events, and instead uses the fixed of the initial window; a vertical scrollbar appears.

When switching reader mode, it's slightly different, but still bothersome: When returning to Webtoon mode, the chapter position is somewhere completely different, and pages don't load, until a chapter is selected in the selector.

The behavior can also be provoked by scrolling over still-loading pages, but it's more finicky and I have not been able to reproduce that reliably.

### Video
https://github.com/user-attachments/assets/1aebe3c4-50ef-428c-89a8-39042c7481ad

cpiber:
It was removed though?

PR #905, opened 2025-03-23
Reader: Reset bounds on window size change

cpiber:
This attempts to handle page resizing a little better than previously (ref 19f2d99).

It's not perfect, but not changing position at all seems impossible due
to how resizes are handled by the browser (since we adjust our content
to the window size).

Partially addresses #902

### DIFF
--- a/src/modules/reader/components/viewer/ReaderChapterViewer.tsx
+++ b/src/modules/reader/components/viewer/ReaderChapterViewer.tsx
@@ -153,6 +153,7 @@ const BaseReaderChapterViewer = ({
     const isCurrentChapterRef = useRef(isCurrentChapter);
     const imageRefs = useRef<(HTMLElement | null)[]>(pages.map(() => null));
     const pagerRef = useRef<HTMLDivElement>(null);
+    const readerNavBarWidthRef = useRef<number>(readerNavBarWidth);
 
     const actualPages = useMemo(() => {
         const arePagesLoaded = !!totalPages;
@@ -261,6 +262,13 @@ const BaseReaderChapterViewer = ({
         window.addEventListener('resize', f);
         return () => window.removeEventListener('resize', f);
     }, [onSizeReset, ref.current]);
+
+    useEffect(() => {
+        if (readerNavBarWidthRef.current === readerNavBarWidth) return;
+        // setting to 0x0 resets the entire state, so on next render the resize observer will re-init with the correct size
+        onSizeReset(0, 0);
+        readerNavBarWidthRef.current = readerNavBarWidth;
+    }, [onSizeReset, readerNavBarWidth]);
 
     const updateState = <T,>(
         value: T,
--- a/src/modules/reader/components/viewer/ReaderChapterViewer.tsx
+++ b/src/modules/reader/components/viewer/ReaderChapterViewer.tsx
@@ -89,6 +89,7 @@ const BaseReaderChapterViewer = ({
     scrollbarYSize,
     readerNavBarWidth,
     onSizeChange,
+    onSizeReset,
     minWidth,
     minHeight,
 }: Pick<
@@ -126,6 +127,7 @@ const BaseReaderChapterViewer = ({
         scrollIntoView: boolean;
         resumeMode: ReaderResumeMode;
         onSizeChange: (width: number, height: number) => void;
+        onSizeReset: (width: number, height: number) => void;
         minWidth: number;
         minHeight: number;
     }) => {
@@ -248,6 +250,17 @@ const BaseReaderChapterViewer = ({
             [onSizeChange, ref.current],
         ),
     );
+
+    useEffect(() => {
+        const f = () => {
+            if (!ref.current || !ref.current.parentElement) return;
+            const { clientWidth, clientHeight } = ref.current.parentElement;
+            onSizeReset(clientWidth, clientHeight);
+        };
+
+        window.addEventListener('resize', f);
+        return () => window.removeEventListener('resize', f);
+    }, [onSizeReset, ref.current]);
 
     const updateState = <T,>(
         value: T,
--- a/src/modules/reader/components/viewer/ReaderViewer.tsx
+++ b/src/modules/reader/components/viewer/ReaderViewer.tsx
@@ -231,6 +231,7 @@ const BaseReaderViewer = forwardRef(
             chapterViewerSize.current.minChapterViewWidth = width;
             chapterViewerSize.current.minChapterViewHeight = height;
             setTriggerReRender({});
+            setPageToScrollToIndex(currentPageIndex);
         };
 
         useReaderHandlePageSelection(
--- a/src/modules/reader/components/viewer/ReaderViewer.tsx
+++ b/src/modules/reader/components/viewer/ReaderViewer.tsx
@@ -227,6 +227,12 @@ const BaseReaderViewer = forwardRef(
             [isContinuousReadingModeActive, isContinuousVerticalReadingModeActive],
         );
 
+        const onChapterViewSizeReset = (width: number, height: number) => {
+            chapterViewerSize.current.minChapterViewWidth = width;
+            chapterViewerSize.current.minChapterViewHeight = height;
+            setTriggerReRender({});
+        };
+
         useReaderHandlePageSelection(
             pageToScrollToIndex,
             currentPageIndex,
@@ -397,6 +403,7 @@ const BaseReaderViewer = forwardRef(
                             scrollbarYSize={scrollbarYSize}
                             readerNavBarWidth={readerNavBarWidth}
                             onSizeChange={onChapterViewSizeChange}
+                            onSizeReset={onChapterViewSizeReset}
                             minWidth={minChapterViewWidth}
                             minHeight={minChapterViewHeight}
                         />


src/modules/reader/components/viewer/ReaderChapterViewer.tsx
```diff
@@ -249,6 +252,24 @@ const BaseReaderChapterViewer = ({
         ),
     );
 
+    useEffect(() => {
```
schroda, line 255: I would also put the `resize` and `readerNavBarWidth` handling for preserving the current page in `useReaderPreserveScrollPosition`, same as in the other pr

src/modules/reader/components/viewer/ReaderChapterViewer.tsx
```diff
@@ -249,6 +252,24 @@ const BaseReaderChapterViewer = ({
...
+        const f = () => {
+            if (!ref.current || !ref.current.parentElement) return;
+            const { clientWidth, clientHeight } = ref.current.parentElement;
+            onSizeReset(clientWidth, clientHeight);
```
schroda, line 259: instead of introducing this function I think it would be cleaner to add the `chapterId`, to which the dimensions belong to, to `chapterViewerSize`
then in `onChapterViewSizeChange` the dimensions should just get overwritten in case they are for the same `chapterId`, this would then handle the case were the width decreases without setting it to 0 and waiting for the resize observer to reset it to the correct size again

cpiber: The problem is that the `ResizeObserver` observes the same element as where the `min-width` is applied. This means that the observed element can *never* get smaller, unless we reset the `min-width`. The usual trick around this is that `min-width` is applied on the parent. I tried this, and this revealed an issue: Many places in the reader use `vw`, which does *not* work when the reader does not occupy the whole viewport (i.e. with static navigation). I am unsure how to proceed; should I attempt to find some solution to `vw`? Just changing to `%` does not work, because the images then have a `max-width` of `0` in Fit Width mode.

Actually, it feels like quite a few of the styles need to be checked again to see when they should apply. For example, setting a `min-width` seems counter-intuitive, when the mode is Fit Width.

schroda: oh, didn't think about that
the problem is that with `infinite scroll` it wasn't possible to style the pages anymore using `%` because that caused issues with the continuous reading modes

what about only setting the `min-width/height` for `ReaderChapterViewers` that aren't the largest, that would be possible with the added `chapterId`

cpiber: I actually managed to resolve that, there was just one place where the navbar width was not properly considered in some situations, leading to this problem. This works quite nicely.

I noticed that preserving the scroll position is only relevant in the Fit modes (Fit Width and Fit Height). However, for both of these it's not enough to do this as a special case when the page is the same, but rather only in response to the two given scenarios (where an actual resize happens), so I still need that function, but it's still cleaner.

cpiber: I'm not entirely convinced all the width values in `getImagePlaceholderStyling` (`ReaderPager.utils.tsx`) are correct, since only `100vw` deducts the navbar width, but it seems to only really be relevant for `fullWidth`, all others are so fleeting that it seems they at most create a scrollbar briefly. Still, intuitively I would say everywhere `vw` is used, the navbar width should be deducted, since that is the available reader space. So I guess in `getMaxWidth` remove the `if` and do it always?

schroda: from a quick look it doesn't seem that the min width/height changes are working as expected, the problem they are intended to fix is appearing again (keeping the different `ReaderChapterViewers` properly centered)

regardless of this, if you want to make changes in that regard I'd prefer if that was done in a separate pr since that is not really related to this one, as well as all your commits will get squashed

regarding
> what about only setting the min-width/height for ReaderChapterViewers that aren't the largest, that would be possible with the added chapterId

if you change `onChapterViewSizeChange` (which seems broken anyway since it's accessing outdated values with `minChapterViewWidth/Height`) to
```
        const onChapterViewSizeChange = useCallback(
            (width: number, height: number, chapterId: ChapterIdInfo['id']) => {
                if (!isContinuousReadingModeActive) {
                    return;
                }

                const isSameChapterId = chapterId === minChapterSizeSourceChapterId;

                if (isContinuousVerticalReadingModeActive) {
                    if (!isSameChapterId && minWidth >= width) {
                      return;
                    }

                    // convert chapterViewerSize to state, since it doesn't make sense for it to be a ref
                    setChapterViewerSize(state => ({
                      width,
                      height: 0,
                      sourceChapterId: chapterId,
                    }))
                    return;
                }

                if (isSameChapterId || minHeight < height) {
                  setChapterViewerSize(state => ({
                    width: 0,
                    height,
                    sourceChapterId: chapterId,
                  }))
                }
            },
            [isContinuousReadingModeActive, isContinuousVerticalReadingModeActive, minChapterViewWidth, minChapterViewHeight, minChapterSizeSourceChapterId],
        );
```

and always pass `0/undefined` for `min width/height` for the `ReaderChapterViewer` of the `ChapterViewerSize source chapter` then the `min width/height` will get updated when the size decreases

cpiber: Maybe I don't quite understand how this is intended to work, but it seems to work fine for me. When a chapter appears with wider images, the reader is rescaled so the previous chapter with narrower pages is still centered.

But your effect is much cleaner, I'll apply that in a moment. Also noticed that the "no previous/next chapter" markers are broken if the reader is smaller than their content, but that's for another PR...

schroda: here is a video which shows the problem

the pages from `chapter 109` are not centered, they don't align with the ones from `chapter 108`

https://github.com/user-attachments/assets/7f8582a5-8098-4791-8ccb-2ec72bbaaa10

cpiber: I can't reproduce that. I created a blank test manga for myself, and it works there (though it's really not pretty).


https://github.com/user-attachments/assets/bea071c9-7524-4faa-85fd-94cf3305433b


https://github.com/user-attachments/assets/308e2db6-f190-46b0-a08b-f5e756bad09b

If you were using adbad6e for that, please quickly pull again, I messed up the effect there and it did not actually store the detected width. Can you also check that the image there is actually smaller and not just black on one side?

schroda: it's still slightly off and it also causes a double vertical scrollbar

I think it be best to just leave it as it is (setting `min width/height` in the `ReaderChapterViewer` for this pr and if you want you can take a look at it in another pr

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -184,8 +185,12 @@ const BaseReaderViewer = forwardRef(
         const handleClick = ReaderControls.useHandleClick(scrollElementRef.current);
 
         const imageRefs = useRef<(HTMLElement | null)[]>(pages.map(() => null));
-        const chapterViewerSize = useRef({ minChapterViewWidth: 0, minChapterViewHeight: 0 });
-        const { minChapterViewWidth, minChapterViewHeight } = chapterViewerSize.current;
+        const [chapterViewerSize, setChapterViewerSize] = useState({
+            minChapterViewWidth: 0,
+            minChapterViewHeight: 0,
+            minChapterSizeSourceChapterId: null as number | null,
```
schroda, line 191:
```suggestion
            minChapterSizeSourceChapterId: -1,
```

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -184,8 +185,12 @@ const BaseReaderViewer = forwardRef(
...
+        });
+        const { minChapterViewWidth, minChapterViewHeight, minChapterSizeSourceChapterId } = chapterViewerSize;
 
         const [, setTriggerReRender] = useState({});
```
schroda, line 195: not needed anymore with `chapterViewerSize` being a state, can be removed

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -209,24 +214,50 @@ const BaseReaderViewer = forwardRef(
...
-                    chapterViewerSize.current.minChapterViewWidth = width;
-                    setTriggerReRender({});
+                const isSameChapterId = chapterId === minChapterSizeSourceChapterId;
+
+                if (isContinuousVerticalReadingModeActive) {
+                    if (!isSameChapterId && minChapterViewWidth >= width) {
+                        return;
+                    }
+
+                    // convert chapterViewerSize to state, since it doesn't make sense for it to be a ref
```
schroda, line 229:
```suggestion
```

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -264,10 +295,16 @@ const BaseReaderViewer = forwardRef(
             visibleChapters,
             isContinuousReadingModeActive,
             readingDirection,
+            readerNavBarWidth,
+            onChapterViewSizeReset,
```
schroda, line 299: you can pass `setPageToScrollToIndex` and the `currentPageIndex` instead, then the whole related logic would be removed from `ReaderViewer`

src/modules/reader/components/viewer/ReaderChapterViewer.tsx
cpiber: Done. I'm fine with this solution. Though I really cannot reproduce your problem. As for the scrollbar, I think that might be fixed by #908

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -264,10 +295,16 @@ const BaseReaderViewer = forwardRef(
...
         );
 
         useLayoutEffect(() => {
-            chapterViewerSize.current = { minChapterViewWidth: 0, minChapterViewHeight: 0 };
+            setChapterViewerSize({
+                minChapterViewWidth: 0,
+                minChapterViewHeight: 0,
+                minChapterSizeSourceChapterId: null,
```
schroda, line 306:
```suggestion
                minChapterSizeSourceChapterId: -1,
```

src/modules/reader/components/viewer/ReaderViewer.tsx
```diff
@@ -397,8 +434,8 @@ const BaseReaderViewer = forwardRef(
                             scrollbarYSize={scrollbarYSize}
                             readerNavBarWidth={readerNavBarWidth}
                             onSizeChange={onChapterViewSizeChange}
-                            minWidth={minChapterViewWidth}
-                            minHeight={minChapterViewHeight}
+                            minWidth={chapter.id === minChapterSizeSourceChapterId ? 0 : minChapterViewWidth}
+                            minHeight={chapter.id === minChapterSizeSourceChapterId ? 0 : minChapterViewHeight}
```
schroda, line 438:
```suggestion
                            minWidth={isChapterSizeSourceChapter  ? 0 : minChapterViewWidth}
                            minHeight={isChapterSizeSourceChapter  ? 0 : minChapterViewHeight}
```

src/modules/reader/utils/ReaderPager.utils.tsx
```diff
@@ -83,7 +83,7 @@ export const getImagePlaceholderStyling = (
         getMaxWidth(`${coerceIn(readerWidthValue ?? width, Math.min(width, readerWidthValue ?? OVER_9000), width)}vw`);
 
     const setReaderWidth = getSetReaderWidth(readerWidth, pageScaleMode);
-    const fullWidth = getMaxWidth(`${Math.min(100, setReaderWidth ?? OVER_9000)}vw`);
+    const fullWidth = `calc(${Math.min(100, setReaderWidth ?? OVER_9000)}vw - ${widthOffset}px)`;
```
schroda, line 86: this should also be reverted right?

cpiber: Sure, it's not as important with `min-width` reverted

schroda (approved):

schroda:
@cpiber you'll have to fix the merge conflicts

cpiber:
@schroda should be good

PR #907, opened 2025-03-23
Basic modal for switching extension settings

### LINKED ISSUE #903: [Feature Request] Make extension settings more obvious
## What feature should be added to Suwayomi?
Not a new feature per-se, more an enhancement.

It is currently possible to configure extensions by browsing latest/popular on the source. Then the settings cog appears in the top bar. This is extremely unintuitive, especially when a source requires configuration before it can be meaningfully used. I actually thought Suwayomi doesn't support configuring the extensions before I noticed the toggle.

Tachiyomi/Mihon shows the settings button in the list of installed extensions, where it is much more obvious and intuitive, and is where most users would expect to be able to configure the extension, though of course this requires introduction of a submenu per-language like in Mihon.

cpiber:
Introduce a small modal to open settings for extensions per source, similar to Mihon.
Since Suwayomi does not support disabling languages, if the source is not configurable the entry in the modal is simply empty.

In the extensions tab, a settings icon is now shown (for installed, upgradable and obsolete items):
20250323_15h44m30s_grim

Clicking simply opens a drawer with the supported languages, and an icon to modify. This button leads to the same page as the icon in the browse tab, no special dialog:
20250323_15h45m32s_grim

Many languages, not configurable:
20250323_15h47m00s_grim

Since this switches to a new page for configuration, it stores the selected item in history state; on back, the drawer is simply reopened.

Closes #903


src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const {
+        data,
+        loading: isLoading,
+        error,
+        refetch,
+    } = requestManager.useGetSourceList({ notifyOnNetworkStatusChange: true });
+
+    if (isLoading || !data) {
+        return (
+            <OptionsPanel open={!!extensionId} onClose={closeDialog}>
```
schroda, line 44: can you use a `Dialog` instead of `OptionsPanel`, I don't plan to keep using that

src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,139 @@
...
+                                    display: 'flex',
+                                    flexDirection: 'column',
+                                    flexGrow: 1,
+                                    flexShrink: 1,
+                                    wordBreak: 'break-word',
+                                    justifyContent: 'center',
+                                }}
+                            >
+                                <Typography variant="h6" component="h3">
+                                    {source.lang}
```
schroda, line 97:
```suggestion
                                   {translateExtensionLanguage(lang)}
```

src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,139 @@
...
+                                >
+                                    {source.isNsfw && (
+                                        <Typography
+                                            variant="caption"
+                                            color="error"
+                                            sx={{
+                                                display: 'inline',
+                                            }}
+                                        >
+                                            {' 18+'}
```
schroda, line 113: I think it's redundant to show the `nsfw` tag here since the extension itself is already marked as such and thus, all sources will be `nsfw` as well

src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const navigate = useNavigate();
+
+    const {
+        data,
+        loading: isLoading,
+        error,
+        refetch,
+    } = requestManager.useGetSourceList({ notifyOnNetworkStatusChange: true });
+
+    if (isLoading || !data) {
```
schroda, line 42:
```suggestion
    if (isLoading) {
```

src/modules/extension/components/ExtensionCard.tsx
```diff
@@ -185,6 +190,21 @@ export function ExtensionCard(props: IProps) {
                         </Typography>
                     )}
                 </Box>
+                {(installedState === InstalledState.UNINSTALL ||
+                    installedState === InstalledState.OBSOLETE ||
+                    installedState === InstalledState.UPDATE) && (
```
schroda, line 195:
```suggestion
                {isInstalled && (
```

src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    if (error) {
+        return (
+            <Dialog open={!!extensionId} onClose={closeDialog}>
+                <EmptyViewAbsoluteCentered
+                    message={t('global.error.label.failed_to_load_data')}
+                    messageExtra={getErrorMessage(error)}
+                    retry={() => refetch().catch(defaultPromiseErrorHandler('Sources::refetch'))}
```
schroda, line 57:
```suggestion
                    retry={() => refetch().catch(defaultPromiseErrorHandler('ExtensionOptions::refetch'))}
```

src/modules/extension/components/ExtensionOptions.tsx
```diff
@@ -0,0 +1,111 @@
...
+                />
+            </Dialog>
+        );
+    }
+
+    const relevantSources = data?.sources.nodes.filter((s) => s.extension.pkgName === extensionId);
+
+    return (
+        <Dialog open={!!extensionId} onClose={closeDialog}>
```
schroda, line 66: the dialogs should have a title (`<DialogTitle>`, you'll probably need to add a new translation to en.json for that) and the content should be wrapped in `<DialogContent>`

probably simpler to just use one dialog and render it like
```
<Dialog ...
  <DialogTitle ...
  <DialogContent>
    {
    (() => {
      if (loading) ...
      if (error) ...
      ...
    })()
    }
   OR
   loading && ...
   !loading && error && ...
   !loading && !error && ...
  </DialogContent>
</Dialog>
```

schroda (approved):

PR #908, opened 2025-03-23
Fixes two places where navbar was not considered

cpiber:
When set to static, the navbar consumes space, so `vw` units are not enough.
On small screens, the `left`+`transform` approach does not work nicely, so just use standard flex alignment, the container is already a flexbox.

As mentioned on #905

### DIFF
--- a/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
+++ b/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
@@ -147,8 +147,11 @@ const BaseReaderTransitionPage = ({
                         ...applyStyles(!isFitWidthPageScaleMode, { alignItems: 'baseline' }),
                     }),
                     ...applyStyles(readingMode === ReadingMode.CONTINUOUS_HORIZONTAL, {
+                        flexDirection: 'row',
+                        maxHeight: `calc(100vh - ${scrollbarXSize}px)`,
+                        position: 'sticky',
+                        top: 0,
                         minWidth: `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px)`,
-                        justifyContent: 'unset',
                     }),
                 }),
             }}
@@ -158,25 +161,12 @@ const BaseReaderTransitionPage = ({
                     gap: 2,
                     maxWidth: (theme) =>
                         // spacing = added padding left + right
-                        `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - ${theme.spacing(2)}px)`,
+                        `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - ${theme.spacing(2)})`,
                     maxHeight: `calc(100vh - ${scrollbarXSize}px)`,
                     width: 'max-content',
                     p: 1,
                     ...applyStyles(isContinuousReadingMode(readingMode), {
-                        ...applyStyles(
-                            // on small screens with "fit to with" enabled, "left 50%" does not center the element in the
-                            // viewport which then causes "translate" to move the element mostly outside the viewport with
-                            // only a small part of it being visible
-                            !isFitWidthPageScaleMode && isContinuousVerticalReadingMode(readingMode),
-                            {
-                                alignSelf: 'center',
-                            },
-                        ),
-                        ...applyStyles(readingMode === ReadingMode.CONTINUOUS_HORIZONTAL, {
-                            position: 'sticky',
-                            top: '50%',
-                            transform: 'translateY(-50%)',
-                        }),
+                        alignSelf: 'center',
                     }),
                 }}
             >
--- a/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
+++ b/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
@@ -156,7 +156,9 @@ const BaseReaderTransitionPage = ({
             <Stack
                 sx={{
                     gap: 2,
-                    maxWidth: `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - 16px)`,
+                    maxWidth: (theme) =>
+                        // spacing = added padding left + right
+                        `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - ${theme.spacing(2)}px)`,
                     maxHeight: `calc(100vh - ${scrollbarXSize}px)`,
                     width: 'max-content',
                     p: 1,
--- a/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
+++ b/src/modules/reader/components/viewer/ReaderTransitionPage.tsx
@@ -140,6 +140,9 @@ const BaseReaderTransitionPage = ({
                     position: 'relative',
                     transform: 'scale(1)',
                     ...applyStyles(isContinuousVerticalReadingMode(readingMode), {
+                        maxWidth: `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px)`,
+                        position: 'sticky',
+                        left: 0,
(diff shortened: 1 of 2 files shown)


src/modules/reader/components/viewer/ReaderTransitionPage.tsx
```diff
@@ -153,7 +153,7 @@ const BaseReaderTransitionPage = ({
             <Stack
                 sx={{
                     gap: 2,
-                    maxWidth: `calc(100vw - ${scrollbarYSize}px)`,
+                    maxWidth: `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - 16px)`,
```
schroda, line 156: what are the `16px`?

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
```diff
@@ -165,8 +165,7 @@ const BaseReaderTransitionPage = ({
                             // only a small part of it being visible
                             !isFitWidthPageScaleMode && isContinuousVerticalReadingMode(readingMode),
                             {
-                                left: '50%',
-                                transform: 'translateX(-50%)',
+                                alignSelf: 'center',
```
schroda, line 168: without the old styling the `transition page` is not always centered in the viewport
if the reader has wide pages and thus a horizontal scrollbar, based on the horizontal scroll position you won't see the  `transition page` anymore

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
cpiber: Padding. I couldn't get it to work properly without subtracting manually from here that

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
cpiber: I see now, I had implemented this because the attempt of moving `min-width` in #905 meant this needed to be done, since also the transition page was now wider. I'll test again

cpiber: Well, this is the old solution on small viewports:

20250324_08h02m08s_grim

This is why I decided to switch. The new solution uses `position: sticky` (so it works with large pages) on the parent element, so centering is still done properly with `alignSelf` (so it works with small viewports)

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
cpiber: Removed, since with the new solution this does not seem to be applied anymore. May have been a bug in Firefox, I added it because frustratingly there was space around in Fit Width mode when there shouldn't be, resulting in a scroll bar.

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
schroda: now it's completly broken in continuous vertical

Screenshot 2025-03-24 193939

cpiber: What browser are you using? It's fine in both Chromium and Firefox...

20250324_19h52m40s_grim

cpiber: Regarding your comment on layout: It does matter when the chapter name is very long. Not all sources write simply "Chapter 108", quite a few give an actual title, which can approach page size

schroda: > What browser are you using? It's fine in both Chromium and Firefox...

Microsoft Edge Version 134.0.3124.85
Firefox 136.0.2

> Regarding your comment on layout: It does matter when the chapter name is very long. Not all sources write simply "Chapter 108", quite a few give an actual title, which can approach page size

can you show a screenshot of what doesn't work, because when I increase e.g. the chapter title I don't see anything wrong

Screenshot 2025-03-24 200339

cpiber: Interesting, this is what it looks like for me:

20250324_20h09m20s_grim

Screenshot taken on Firefox, but same problem on Chromium, will try to test on a Windows system shortly

schroda: ignore me, I had unreverted local changes... my bad

schroda: > Interesting, this is what it looks like for me:

looks like this only happens on larger screens

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
```diff
@@ -153,23 +156,22 @@ const BaseReaderTransitionPage = ({
             <Stack
                 sx={{
                     gap: 2,
-                    maxWidth: `calc(100vw - ${scrollbarYSize}px)`,
+                    maxWidth: `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - 16px)`,
```
schroda, line 159:
```suggestion
                    maxWidth: (theme) =>
                        // spacing = added padding left + right
                        `calc(100vw - ${scrollbarYSize}px - ${readerNavBarWidth}px - ${theme.spacing(2)}px)`,
```

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
cpiber: Yeah, it's weird, as I make the window bigger, the text slides more to the left...

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
```diff
@@ -153,23 +156,22 @@ const BaseReaderTransitionPage = ({
...
-                        position: 'sticky',
                         ...applyStyles(
                             // on small screens with "fit to with" enabled, "left 50%" does not center the element in the
                             // viewport which then causes "translate" to move the element mostly outside the viewport with
                             // only a small part of it being visible
                             !isFitWidthPageScaleMode && isContinuousVerticalReadingMode(readingMode),
                             {
-                                left: '50%',
-                                transform: 'translateX(-50%)',
+                                alignSelf: 'center',
```
schroda, line 170: is `!isFitWidthPageScaleMode` and the comment still relevant with your changes or can that be removed?

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
```diff
@@ -153,23 +156,22 @@ const BaseReaderTransitionPage = ({
...
                             },
                         ),
                         ...applyStyles(readingMode === ReadingMode.CONTINUOUS_HORIZONTAL, {
```
schroda, line 173: could the horizontal reader also be changed like this? I'd assume both horizontal and vertical would then share more code and the different handling could be removed/reduced?

cpiber: If the flex mode is changed to `row` for that mode (it's currently always `column`, but where it's set I'd have to check), and the same `max-height` limitation is applied to the parent element, then yes, it should work, I'll try in a moment

cpiber: Good idea, cleans up this part quite nicely

src/modules/reader/components/viewer/ReaderTransitionPage.tsx
cpiber: You are right, removed it, this is now fixed

schroda (approved):

cpiber:
Thanks for your patience and insight on this and my other PRs :)

schroda:
thank you for putting in all the work 🙂

PR #913, opened 2025-03-30
Update scroll preserve by tracking cursor

### LINKED ISSUE #912: Scrolling to previous chapter in continuous modes jumps to next
### Steps to reproduce

1. Open any chapter that isn't the first
2. Switch to any continuous mode (Continuous Vertical in video)
3. Scroll to previous chapter

### Expected behavior

See the last page of the previous chapter

### Actual behavior

The transition page to the previous chapter is shown, then it jumps briefly to the transition page to the next chapter, then switches to the first page of the next chapter. Scrolling up again shows the transition page to the next (now current) chapter again.

This only happens if the previous chapter is not loaded, i.e. it's possible to now scroll up properly.

### Suwayomi-WebUI version

r2415 (master)

### Suwayomi-Server version

v1.1.1-r1716

### Used web browser

Firefox 136.0.4

### Client operating system

Ubuntu 24.04.2 LTS

### Server operating system

Ubuntu 24.04.2 LTS

### Server JVM version

java version "21.0.6" 2025-01-21 LTS

### Other details

https://github.com/user-attachments/assets/34c7cc82-f0eb-4eee-b6ce-b663189a1a24

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have checked the [ongoing preview changelog](https://github.com/Suwayomi/Suwayomi-WebUI/issues/749) and this bug has **NOT** been listed as fixed
- [x] This bug is UI related and does not belong to the server. If in doubt, create the request in the [Suwayomi-Server repository](https://github.com/Suwayomi/Suwayomi-Server/issues).
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have gone through the [Suwayomi-Server wiki](https://github.com/Suwayomi/Suwayomi-Server/wiki) and [Discord FAQ and support](https://discord.gg/DDZdqZWaHA).
- [x] I have updated the Suwayomi-WebUI and Suwayomi-Server to the latest versions (**[Suwayomi-WebUI](https://github.com/suwayomi/suwayomi-webui/releases/latest)**, **[Suwayomi-Server](https://github.com/suwayomi/suwayomi-server/releases/latest)**).
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Suwayomi does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

cpiber:
Closes #912

This PR tracks which element was (roughly) in focus by the user depending on scroll position. Then when elements are added, it attempts to scroll back to the old position by using the new position of the focused element. In effect, this allows to more closely estimate if new content was added in front or behind the current reading position.

This is not perfect. The video below shows what happens most of the time; scroll position is reset to where it should be. Sometimes, the user instead ends up on the last or second to last page, not sure why that happens (possibly different order of image loads?)

https://github.com/user-attachments/assets/7cccb707-10db-4c1d-92aa-5767de58b670

Tested in all reading modes. I have not seen any bad side effects in other situations; it does help a bit with forward scrolling during load, but not as much as in this situation.

### DIFF
--- a/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
+++ b/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
@@ -75,6 +75,19 @@ export const useReaderPreserveScrollPosition = (
             return () => {};
         }
 
+        const updateObservation = (
+            nodes: NodeList,
+            resizeAction: (n: Element) => void,
+            intersectionAction: (n: Element) => void,
+        ) =>
+            Array.from(nodes)
+                .filter((n) => n instanceof HTMLElement)
+                .flatMap((n) => {
+                    resizeAction(n);
+                    return Array.from(n.querySelectorAll('img'));
+                })
+                .forEach(intersectionAction);
+
         const resizeObserver = new ResizeObserver(onDoPreserveScroll);
         const intersectionObserver = new IntersectionObserver((entries) => {
             // find the first visible image inside the viewport
@@ -89,22 +102,16 @@ export const useReaderPreserveScrollPosition = (
         });
         const mutationObserver = new MutationObserver((entries) => {
             for (const entry of entries) {
-                for (const added of entry.addedNodes) {
-                    if (added instanceof HTMLElement) {
-                        resizeObserver.observe(added);
-                        for (const img of added.querySelectorAll('img')) {
-                            intersectionObserver.observe(img);
-                        }
-                    }
-                }
-                for (const removed of entry.removedNodes) {
-                    if (removed instanceof HTMLElement) {
-                        resizeObserver.unobserve(removed);
-                        for (const img of removed.querySelectorAll('img')) {
-                            intersectionObserver.unobserve(img);
-                        }
-                    }
-                }
+                updateObservation(
+                    entry.addedNodes,
+                    (n) => resizeObserver.observe(n),
+                    (n) => intersectionObserver.observe(n),
+                );
+                updateObservation(
+                    entry.removedNodes,
+                    (n) => resizeObserver.unobserve(n),
+                    (n) => intersectionObserver.unobserve(n),
+                );
             }
         });
         mutationObserver.observe(element, {
--- a/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
+++ b/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
@@ -7,14 +7,7 @@
  */
 
 import { RefObject, useCallback, useEffect, useLayoutEffect, useRef } from 'react';
-import { ChapterIdInfo } from '@/modules/chapter/services/Chapters.ts';
-import {
-    ReaderPageScaleMode,
-    ReaderStateChapters,
-    ReadingDirection,
-    ReadingMode,
-} from '@/modules/reader/types/Reader.types.ts';
-import { TChapterReader } from '@/modules/chapter/Chapter.types.ts';
+import { ReaderPageScaleMode, ReadingMode } from '@/modules/reader/types/Reader.types.ts';
 import { ReaderStatePages } from '@/modules/reader/types/ReaderProgressBar.types';
 import { isReaderWidthEditable } from '@/modules/reader/utils/ReaderSettings.utils.tsx';
 
@@ -35,14 +28,9 @@ const getActiveScrollElement = (root: HTMLElement): HTMLElement | undefined => {
 
 export const useReaderPreserveScrollPosition = (
     scrollElementRef: RefObject<HTMLElement | null>,
-    currentChapterId: ChapterIdInfo['id'] | undefined,
-    chapterIndex: number,
     pageIndex: number,
-    chaptersToRender: TChapterReader[],
-    visibleChapters: ReaderStateChapters['visibleChapters'],
     readingMode: ReadingMode,
     isContinuousReadingModeActive: boolean,
-    readingDirection: ReadingDirection,
     readerNavBarWidth: number,
     setPageToScrollToIndex: ReaderStatePages['setPageToScrollToIndex'],
     pageScaleMode: ReaderPageScaleMode,
--- a/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
+++ b/src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
@@ -96,16 +96,12 @@ export const useReaderPreserveScrollPosition = (
             for (const entry of entries) {
                 for (const added of entry.addedNodes) {
                     if (added instanceof HTMLElement) {
-                        for (const img of added.querySelectorAll('img')) {
-                            resizeObserver.observe(img);
-                        }
+                        resizeObserver.observe(added);
                     }
                 }
                 for (const removed of entry.removedNodes) {
                     if (removed instanceof HTMLElement) {
-                        for (const img of removed.querySelectorAll('img')) {
-                            resizeObserver.unobserve(img);
-                        }
+                        resizeObserver.unobserve(removed);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
```diff
@@ -59,52 +69,57 @@ export const useReaderPreserveScrollPosition = (
...
 
-        scrollElement.scrollTo(newLeft, newTop);
-    }, [currentChapterId]);
+        const resizeObserver = new ResizeObserver(onDoPreserveScroll);
+        const mutationObserver = new MutationObserver((entries) => {
+            for (const entry of entries) {
+                for (const added of entry.addedNodes) {
+                    if (added instanceof HTMLElement) {
+                        for (const img of added.querySelectorAll('img')) {
+                            resizeObserver.observe(img);
```
schroda, line 100: would it be enough to only observe the `ReaderChapterViewer` elements (so the direct children of the `scroll element`), since they also change in `width/height` in case their children change?

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
```diff
@@ -7,35 +7,41 @@
...
-} from '@/modules/reader/types/Reader.types.ts';
-import { getOptionForDirection } from '@/modules/theme/services/ThemeCreator.ts';
-import { READING_DIRECTION_TO_THEME_DIRECTION } from '@/modules/reader/constants/ReaderSettings.constants.tsx';
-import { getPreviousNextChapterVisibility } from '@/modules/reader/utils/Reader.utils.ts';
-import { TChapterReader } from '@/modules/chapter/Chapter.types.ts';
+import { ReaderPageScaleMode, ReadingMode } from '@/modules/reader/types/Reader.types.ts';
 import { ReaderStatePages } from '@/modules/reader/types/ReaderProgressBar.types';
 import { isReaderWidthEditable } from '@/modules/reader/utils/ReaderSettings.utils.tsx';
 
+const getActiveScrollElement = (root: HTMLElement): HTMLElement | undefined => {
```
schroda, line 14: I feel like this function is too hard to read and understand, can it be cleaned up?

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
cpiber: I tried that briefly, and it didn't work, don't remember why though, I'll look into it

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
cpiber: What is hard to read? I feel it's pretty straightforward, it finds the first element inside the visible area. I could add some comments?

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
```diff
@@ -59,52 +69,57 @@ export const useReaderPreserveScrollPosition = (
...
-        const newTop = wasScrolledBackwardVertical ? heightOfPrependedContent + top : top;
-
-        // only relevant when prepending content to the dom due to the resulting layout shift
-        const isFirstRenderOfPreviousChapter =
-            !previousNextChapterVisibility.previous && (wasScrolledBackwardHorizontal || wasScrolledBackwardVertical);
-        if (!isFirstRenderOfPreviousChapter) {
-            return;
+        const newLeft = left - activeLeft + active.offsetLeft;
+        const newTop = top - activeTop + active.offsetTop;
+        scrollElement.scrollTo(newLeft, newTop);
```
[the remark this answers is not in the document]
cpiber: It should, though admittedly I didn't realize that there's a right to left mode, I will test this again.
I don't see a problem with it, since this simply re-orients according to the active element, but `getActiveScrollElement` might need to be changed a little for that.

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
schroda: > it finds the first element inside the visible area

the function name should represent this somehow (`getFirstVisibleElement` or something like that)
`activeScrollElement` sounds like it is the element that handles the scrolling

> What is hard to read

the loop in combination with the mutation thats going on, I think making it `recursive` should already improve it a lot

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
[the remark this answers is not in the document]
schroda: just to make sure, I had some issues with that because the scroll position will be negative (ltr: `0 - scroll width` -> rtl: `-(scroll width) - 0`)

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
```diff
@@ -46,10 +52,14 @@ export const useReaderPreserveScrollPosition = (
         }
 
         const onScroll = () => {
+            const active = getActiveScrollElement(element);
```
schroda, line 55: won't this be a problem?
this will run on every scroll event and potentially could loop over > 1000 elements on each event

schroda: I think instead of this, using the `intersection observer` to get the first visible element would be a better option

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
cpiber: > I think making it recursive should already improve it a lot

Funny, most people would disagree and say that recursive functions are more difficult, but I'll have a look in the evening

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
cpiber: The problem with the intersection observer is that it's possible for multiple images to be visible at the same time. Especially when scrolling forward, this could lead to unintuitive behavior, but I'll test it.

cpiber: Only got to it today. I implemented it like that; this now tracks *any* element inside the viewport, and due to how scrolling works, this is likely the last element the user scrolled to, instead of the first visible. In my testing, this did not cause any problems though, since all the images on screen are usually loaded anyway, so which we use as anchor does not matter.

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
schroda: only observing the `childList` with the mutation observer seems to work as expected

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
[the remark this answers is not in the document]
schroda: > RTL loading is currently broken:

should be fixed with the latest preview version

> How would I go about testing Arabic? I haven't seen that language in the picker. I tried setting my browser's content language to ar-EG, which made Suwayomi render dates in arabic, but I haven't observed anything else.

I checked it and it's working

looks like the language got removed after I've added the requirement for languages to at least be 75% translated, you would have to add the language manually to the available resources list (`i18n/index.ts -> i18nResources`)

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
cpiber: Yep, see updated code

src/modules/reader/hooks/useReaderPreserveScrollPosition.ts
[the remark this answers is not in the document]
cpiber: Good to know, thanks for checking

schroda (approved):

PR #920, opened 2025-04-08
fix VersionMapping for people on older preview versions

Robonau:
fix VersionMapping for people on older preview versions

fixes the issue this guy is having https://discord.com/channels/801021177333940224/1339229198434439230/1359081114693795850


schroda:
since there will be a server release soon I don't want to re-add this for some odd case of someone using some old ass server preview version

there is also always the option to use an older preview version with the custom flavor

PR #940, opened 2025-05-16
Add author/artist sort

Well2333:
Added the ability to sort manga by artist/author.

Key changes:
- Added new sort mode `byArtistOrAuthor`
- Added proper translations for all supported languages

添加了按作者/画师排序漫画的功能。 

主要修改：
- 新增 `byArtistOrAuthor` 排序模式
- 为所有支持的语言添加对应翻译

image


.gitignore
```diff
@@ -8,4 +8,5 @@ build/*
 
 tools/scripts/tokens.json
 
-src/lib/graphql/schema.json
\ No newline at end of file
+src/lib/graphql/schema.json
+pnpm-lock.yaml
```
schroda, line 12: this should not be added here

public/locales/de.json
```diff
@@ -10,7 +10,8 @@
                     "by_date_added": "Kürzlich hinzugefügt",
                     "by_latest_uploaded_chapter": "Zuletzt hochgeladenes Kapitel",
                     "by_latest_fetched_chapter": "Zuletzt abgerufenes Kapitel",
-                    "by_total_chapters": "Gesamte Anzahl der Kapitel"
+                    "by_total_chapters": "Gesamte Anzahl der Kapitel",
```
schroda, line 13: only `en` gets translated in the repo, all other languages are translated in `weblate`

src/modules/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -193,6 +201,43 @@ const sortManga = <Manga extends TMangaSort>(
         case 'totalChapters':
             result.sort((a, b) => sortByNumber(a.chapters.totalCount, b.chapters.totalCount));
             break;
+        case 'byArtistOrAuthor':
```
schroda, line 204: This should stay consistent with the other sorting options.
Which means that it should be split into two seperate options as well as it should only sort the artist/author value and nothing else as a fallback

Well2333: well. in my case most of the  gallery only have one of the artists or author. can I keep zhe merged options (sort by author, artist, author/artist 3 types of sorting)?

schroda: No, I don't want to add something like that, for the already mentioned reason.
If you can't live without that, then you have the option to host your custom build instead of using the default one.

Well2333: I believe that the `author` and `artist` fields are inherently closely related in describing creator identity, which fundamentally differs from sorting dimensions purely based on titles, update time, or read/unread statuses.  

Meanwhile, due to differences in data sources or uploader operations, the same creator may be recorded in different fields (e.g., the creator name exists only in either `author` or `artist`, with the other field marked as `Unknown`). This creates critical flaws in single-field classification mechanisms:  
1. Galleries using the alternative field for creator labeling would be categorized under other sorting methods (though solvable by setting secondary sorting priorities)  
2. Works by the same creator become fragmented between two classification nodes: the properly categorized section and the `Unknown` section (unsolvable)  

For instance, when categorizing by `author` only:  
```
G0 => author:A artist:A  
G1 => author:A artist:Unknown  
G2 => author:B artist:Unknown  
G3 => author:C artist:Unknown  
G4 => author:D artist:Unknown  
G5 => author:Unknown artist:E  
G6 => author:Unknown artist:A  
```  
Clearly, this is counterintuitive.  

Therefore, implementing a field fallback mechanism (automatically querying alternative fields when the primary field is empty) can effectively avoid this issue. Hence, I assert that retaining the field fallback mechanism is absolutely necessary.  

---

我认为，`author`与`artist`字段本质上是描述创作者身份的有密切关系的字段，而这与纯粹基于作品名称、更新时间或已读状态等完全无关的排序维度存在本质区别。  

以此同时，根据数据源或上传者操作的差异，同一创作者可能被录入到不同字段中（如作者名仅存在于`author`或`artist`中的单一字段，另一字段为`Unknown`）。这将导致单字段分类机制存在严重缺陷，即当仅依据单一字段进行归类时：  
1. 采用另一字段标注创作者的画廊将按照其他排序方式进行分类（但这可以通过设置次选的排序优先级解决）  
2. 相同创作者的作品会被割裂在两个分类节点——正常归类区与`Unknown`（未识别）区（无法解决）  
例如，以下是仅根据 `author` 进行分类的情况：  
```
G0 => author:A artist:A  
G1 => author:A artist:Unknown  
G2 => author:B artist:Unknown  
G3 => author:C artist:Unknown  
G4 => author:D artist:Unknown  
G5 => author:Unknown artist:E  
G6 => author:Unknown artist:A  
```  
很明显，这显然并不是符合直觉的。  

因此，通过实施字段回退机制（即在主分类字段为空值时自动查询备用字段），能够有效的避免这个问题，所以我认为保留字段回退机制机制是十分有必要的。

schroda: I'm already not seeing the benefit of being able to sort by author or artist.
You aren't able to distinguish which the first/last entry of an artist/author is, so how is it beneficial to sort by this information?
Having this sort option which then just "arbitrarily" sorts your entries, makes this even worse.

So again, no, I don't plan to add such a sort option.

Well2333:
If we split it into two according to the method you mentioned, then this sorting method indeed has no value of existence.

PR #966, opened 2025-06-06
Feat: Adds to Reader settings custom scroll amount slider

FedericoRossiIT:
## What it does
- Added an even slower speed scroll speed `AS MOUSEWHEEL` that is set to 5%
- Added custom scroll speed slider in range `[0.5-100]`
  - The slider use the same variable as the `Scroll amount` buttons so if a button is clicked it update the slider

Example custom value image:
immagine

Example button value image:
immagine

## Why the changes?
I couldn't read the manga as it was before, when reading manga on scanalations i use arrows to scroll and it was scrolling too much of the page.
This is best in continuous or webtoon reading mode.

### DIFF
--- a/src/modules/reader/constants/ReaderSettings.constants.tsx
+++ b/src/modules/reader/constants/ReaderSettings.constants.tsx
@@ -40,7 +40,7 @@ import { WebtoonPageIcon } from '@/assets/icons/svg/WebtoonPageIcon.tsx';
  * percentage values
  */
 export enum ReaderScrollAmount {
-    AS_MOUSEWHEEL = 5,
+    TINY = 10,
     SMALL = 25,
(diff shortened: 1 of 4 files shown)


src/modules/reader/constants/ReaderSettings.constants.tsx
```diff
@@ -40,6 +40,7 @@ import { WebtoonPageIcon } from '@/assets/icons/svg/WebtoonPageIcon.tsx';
  * percentage values
  */
 export enum ReaderScrollAmount {
+    AS_MOUSEWHEEL = 5,
```
schroda, line 43: Rename it to e.g. `Tiny` and make it `10%`, anything below that should be set via the slider

PR #984, opened 2025-07-09
Update server settings for Authentication modes

cpiber:
Screen Shot 2025-07-09 at 20 00 40

In Suwayomi/Suwayomi-Server#1498, a login page as introduced as login mode. For this, the `basicAuthEnabled` setting was migrated to a `authMode` enum-based setting.

I tried adding a tooltip for why the mode setting is disabled, but it never showed up, not sure how to use that in a list (the original setting didn't have one either, so you probably ran to that as well).

Note: The server, in the interest of compatibility, automatically translates reads/writes from/to `basicAuthEnabled` to the `authMode` setting. Same for `basicAuthUsername`/`basicAuthPassword` to/from `authUsername`/`authPassword`. It is thus not critical to update this, for compatibility with stable.


schroda (approved):
> I tried adding a tooltip for why the mode setting is disabled, but it never showed up, not sure how to use that in a list (the original setting didn't have one either, so you probably ran to that as well).

Tbh, I didn't even think about adding a tooltip for that

PR #1003, opened 2025-08-22
[Improvment] Drag scrolling inertia calculation no longer locked to 60hz

bornav:
With the current implementation of the mouse drag inertia, the animation loop was locked to 60hz, making the transition from hold dragging to velocity calculation quite jarring, since it went from high refreshrate, to 60hz

InertiaMove was statically defined to 16ms,  now we use requestAnimationFrame that should have the refreshrate of the browser

### DIFF
--- a/src/base/hooks/useMouseDragScroll.tsx
+++ b/src/base/hooks/useMouseDragScroll.tsx
@@ -37,7 +37,7 @@ export const useMouseDragScroll = (
     const previousClickPosY = useRef<Positions>([0, 0, 0]);
     const previousClickTime = useRef<ClickTimes>([0, 0, 0]);
     const scrollAtT0 = useRef<[ScrollLeft: number, ScrollTop: number]>([0, 0]);
-    const inertiaTimeInterval = useRef<NodeJS.Timeout>(undefined);
+    const inertiaTimeInterval = useRef<NodeJS.Timeout | number>(undefined);
 
     useEffect(() => {
         const element = ref?.current;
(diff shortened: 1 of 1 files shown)


src/base/hooks/useMouseDragScroll.tsx
```diff
@@ -37,7 +37,7 @@ export const useMouseDragScroll = (
     const previousClickPosY = useRef<Positions>([0, 0, 0]);
     const previousClickTime = useRef<ClickTimes>([0, 0, 0]);
     const scrollAtT0 = useRef<[ScrollLeft: number, ScrollTop: number]>([0, 0]);
-    const inertiaTimeInterval = useRef<NodeJS.Timeout>(undefined);
+    const inertiaTimeInterval = useRef<NodeJS.Timeout | number>(undefined);
```
schroda, line 40:
```suggestion
    const inertiaTimeInterval = useRef<number>(undefined);
```

`NodeJS.Timeout` is not possible, since the  `setTimeout` usage was removed

src/base/hooks/useMouseDragScroll.tsx
```diff
@@ -68,7 +68,16 @@ export const useMouseDragScroll = (
...
+        const clearInertiaInterval = () => {
+            if (inertiaTimeInterval.current !== undefined) {
+                if (typeof inertiaTimeInterval.current === 'number') {
+                    cancelAnimationFrame(inertiaTimeInterval.current);
+                } else {
+                    clearInterval(inertiaTimeInterval.current);
+                }
+                inertiaTimeInterval.current = undefined;
+            }
+        };
```
schroda, line 80:
```suggestion
        const clearInertiaInterval = () => {
            if (inertiaTimeInterval.current !== undefined) {
                cancelAnimationFrame(inertiaTimeInterval.current);
                inertiaTimeInterval.current = undefined;
            }
        };
```

schroda (approved):

PR #1051, opened 2026-01-08
Improve library manga sorting

### LINKED ISSUE #1050: Add secondary sort for manga in library
### Describe your suggested feature

There should be a secondary sort for manga in the library. When sorting manga by the selected option, items with the same value should be ordered by name. This is also how it works in Mihon.

### Other details

_No response_

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have checked the [ongoing preview changelog](https://github.com/Suwayomi/Suwayomi-WebUI/issues/749) and this feature is **NOT** yet listed.
- [x] This feature is UI related and does not belong to the server. If in doubt, create the request in the [Suwayomi-Server repository](https://github.com/Suwayomi/Suwayomi-Server/issues).
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated the Suwayomi-WebUI and Suwayomi-Server to the latest versions (**[Suwayomi-WebUI](https://github.com/suwayomi/suwayomi-webui/releases/latest)**, **[Suwayomi-Server](https://github.com/suwayomi/suwayomi-server/releases/latest)**).
- [x] I have filled out all of the requested information in this form, including specific version numbers.

Bartuzen:
Closes #1050 

Made some improvements to library manga sorting, which also makes the behavior match Mihon's. See [here](https://github.com/mihonapp/mihon/blob/bd5c4d48f980d2d3dcc1112fe499dba17ef8e507/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt#L297-L351) for Mihon's implementation.

Changes:
- Sorting by title is now case-insensitive
- Added a secondary sort by title when the primary sort value is the same
- ~~Updated unread chapter sorting so manga with unread chapters are always shown before those without, regardless of sort order~~

### DIFF
--- a/src/features/library/hooks/useGetVisibleLibraryMangas.ts
+++ b/src/features/library/hooks/useGetVisibleLibraryMangas.ts
@@ -183,12 +183,7 @@ const sortManga = <Manga extends TMangaSort>(
             case 'dateAdded':
                 return (a, b) => sortByNumber(a.inLibraryAt, b.inLibraryAt);
             case 'unreadChapters':
-                return (a, b) => {
-                    if (a.unreadCount === b.unreadCount) return 0;
-                    if (a.unreadCount === 0) return desc ? -1 : 1;
-                    if (b.unreadCount === 0) return desc ? 1 : -1;
-                    return sortByNumber(a.unreadCount, b.unreadCount);
-                };
+                return (a, b) => sortByNumber(a.unreadCount, b.unreadCount);
             case 'lastRead':
                 return (a, b) => sortByNumber(a.lastReadChapter?.lastReadAt, b.lastReadChapter?.lastReadAt);
             case 'latestUploadedChapter':


schroda:
I think I prefer keeping the natural sort for unread chapters.

> I personally use unread chapter sorting with descending order in my reading category, so I can quickly see manga I'm not up to date with

In that case the change wouldn't have any effect and

>  I can also imagine someone in the same situation who prefers ascending order, and in that case they likely wouldn't want to scroll to the bottom every time just to find manga they haven't caught up on yet.

here you can just filter the fully read entries out. This actually what I do.
Since filtering and sorting is per category, I also don't see a scenario where you would have to enable and disable the filter often or at all.
The only time I adjust the filters in that category is to check if there are some entries that haven't been updated in a while. In that case I want to have the natural sort order and not have to scroll all the way down to find the fully read entries.

PR #1085, opened 2026-04-12
Chapters grouped by manga series

matsuaa:
This pull request is for the issue#1834
This pr change the history page to only show the most recent chapter read for each manga series, hiding duplicate entries for the same series. I also asked on the discord server for an approval and they agreed on that visual.
image1
image2
Closes the issue #1834


schroda:
You have to handle the case were a load has not enough different manga to be able to trigger the next load on scroll.
See [here](https://github.com/Suwayomi/Suwayomi-WebUI/blob/master/src/features/manga/components/MangaGrid.tsx#L296) how it's done for the manga grid

DraChimeraVee:
Nice, finally the most beautiful J2K feature

schroda:
> You have to handle the case were a load has not enough different manga to be able to trigger the next load on scroll. See [here](https://github.com/Suwayomi/Suwayomi-WebUI/blob/master/src/features/manga/components/MangaGrid.tsx#L296) how it's done for the manga grid

This actually includes [this](https://github.com/Suwayomi/Suwayomi-WebUI/blob/master/src/features/source/browse/screens/SourceMangas.tsx#L396) as well. If the next page only includes chapters of manga that already have been loaded/seen, it's an empty load and the next page needs to be loaded immediately

schroda:
~The conflicts need to be solved~

src/features/history/screens/History.tsx
```diff
@@ -40,7 +42,19 @@ export const History: React.FC = () => {
...
+        const seenMangaIds = new Set<number>();
+        return allReadEntries.filter((chapter) => {
+            if (seenMangaIds.has(chapter.manga.id)) {
+                return false;
+            }
+            seenMangaIds.add(chapter.manga.id);
+            return true;
+        });
+    }, [allReadEntries]);
+    const filteredOutAllItemsOfFetchedPage = allReadEntries.length > 0 && readEntries.length === 0;
```
schroda: `chapterHistoryData` is returning the full history (page 1...n), which means that this won't work for any page request after the initial one.

`filteredOutAllItemsOfFetchedPage` needs to reflect the latest page fetch

schroda (changes requested):
I've checked it locally and it won't load any page after the initial load has completed, i.e. it fetched more pages until the scrollbar appeared.

A next page is available but when scrolling down to the bottom, it does not trigger the next load.
This is most likely connected with how virtuosos `endReached` callback works. It fires once the last item gets rendered

DraChimeraVee:
Any update?

schroda:
Now it skips the requests for 100-150 and 250-300, the last one causes it to spam 300-350 until you leave the history page


Screenshot 2026-05-25 003611
Screenshot 2026-05-25 003618
Screenshot 2026-05-25 003625
Screenshot 2026-05-25 003630
Screenshot 2026-05-25 003638
Screenshot 2026-05-25 003644
Screenshot 2026-05-25 003652

vtorres-t:
`StyledGroupedVirtuoso endReached()` has a bug: it only loads the first time. I've fixed it with `atBottomStateChange()`.

I have another alternative here. https://github.com/vtorres-t/Suwayomi-WebUI/blob/master/src/features/history/screens/History.tsx

image

image

schroda (approved):

PR #1126, opened 2026-07-09
Feature Wake Lock Screen while reading

vtorres-t:
image

New feature to keep the screen on while reading. This can be configured.


src/features/metadata/Metadata.constants.ts
```diff
@@ -329,6 +329,9 @@ export const APP_METADATA: Record<
     shouldInformAboutScanlatorChange: {
         convert: convertToBoolean,
     },
+    shouldKeepScreenReading: {
```
schroda:
```suggestion
    shouldWakeLockScreen: {
```

or something else if you have a better name than the current one

src/features/reader/settings/behaviour/ReaderBehaviourSettings.tsx
```diff
@@ -136,6 +136,11 @@ export const ReaderBehaviourSettings = ({
                 checked={settings.shouldInformAboutScanlatorChange}
                 onChange={(_, checked) => updateSetting('shouldInformAboutScanlatorChange', checked)}
             />
+            <CheckboxInput
+                label={t`Keep screen on while reading`}
```
schroda:
```suggestion
                label={t`Keep screen on`}
```

src/features/reader/wake-lock/useReaderWakeLock.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright (C) Contributors to the Suwayomi project
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+import { useEffect, useRef } from 'react';
+
+export function useReaderWakeLock(isLoading: boolean, shouldKeepScreenReading: boolean) {
```
schroda: Why should it not get locked while it's loading?

src/features/reader/wake-lock/useReaderWakeLock.ts
```diff
@@ -0,0 +1,60 @@
...
+    const wakeLockSentinelRef = useRef<any | null>(null);
+
+    useEffect(() => {
+        const requestWakeLock = async () => {
+            if (!('wakeLock' in navigator)) {
+                return;
+            }
```
schroda:
```suggestion
    if (!('wakeLock' in navigator)) {
        return;
    }

    useEffect(() => {
        const requestWakeLock = async () => {
```

src/features/reader/wake-lock/useReaderWakeLock.ts
```diff
@@ -0,0 +1,60 @@
...
+                }
+
+                wakeLockSentinelRef.current = await navigator.wakeLock.request('screen');
+
+                wakeLockSentinelRef.current.addEventListener('release', () => {
+                    wakeLockSentinelRef.current = null;
+                });
+            } catch (err) {
+                // oxlint-disable-next-line no-console
+                console.error('Error al activar Wake Lock en el lector:', err);
```
schroda:
```suggestion
```

schroda:
The changelog needs to be updated as well

src/features/reader/wake-lock/useReaderWakeLock.ts
vtorres-t: Avoid unnecessary exceptions: If the GraphQL backend is slow to respond, the Wake Lock API might throw errors if invoked repeatedly on a blocked background thread or before the first user interaction.

Visual consistency: Ensure that the screen-on behavior begins exactly when the content is ready for the user to consume.

schroda: > Avoid unnecessary exceptions

The exception don't have any impact and it would get called at most twice due to the settings loading. I don't think this is a problem

> Visual consistency

What visual change is there supposed to be due to wake locking the device?

The lock should just be acquired once the reader gets opened

src/features/reader/wake-lock/useReaderWakeLock.ts
schroda: The hook doesn't need to be called in case the api is not supported

Issue #1005, opened 2025-09-01
Random Library Entry Button / Filter


Natsu972:
### Describe your suggested feature

- A button (e.g., “Random”) in the UI that picks and opens a random manga from the library.
- Alternatively, a filtering/sorting option that can shuffle entries randomly.

Here are two examples from Mihon  :

![Image](https://github.com/user-attachments/assets/ebc390e9-927e-4773-aca8-8a6d38833105)

![Image](https://github.com/user-attachments/assets/1ce81d31-8db7-4e7f-a0ac-e0edac196dc8)

### Other details

The random button should only open entries with at least 1 unread chapter, unlike Mihon, which can open entries without unread chapters.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have checked the [ongoing preview changelog](https://github.com/Suwayomi/Suwayomi-WebUI/issues/749) and this feature is **NOT** yet listed.
- [x] This feature is UI related and does not belong to the server. If in doubt, create the request in the [Suwayomi-Server repository](https://github.com/Suwayomi/Suwayomi-Server/issues).
- [x] I have written a short but informative title (ideally less than ~100 characters).
- [x] I have updated the Suwayomi-WebUI and Suwayomi-Server to the latest versions (**[Suwayomi-WebUI](https://github.com/suwayomi/suwayomi-webui/releases/latest)**, **[Suwayomi-Server](https://github.com/suwayomi/suwayomi-server/releases/latest)**).
- [x] I have filled out all of the requested information in this form, including specific version numbers.

PR #1129, opened 2026-07-17
Added random sorting option for library

themaskedshell:
Adds a random sorting option to the library sort menu. Resolves part of #1005.

Changes:
-Adds a random sort button to the sort UI.
-Flips the filter/sort order to sort then filter.
-Adds replay icon from mui for random button.


schroda:
Instead of showing the sort desc and asc arrow icons for `Random` it should probably be some refresh icon, like it is done in mihon.

In `useGetVisibleLibraryMangas` the sorting should be applied before filtering, so that the items do not get randomized again everytime a filter is changed.

Should these items be randomized again everytime the category gets opened again, even if it was already opened and randomized, what is the expected behavior here?
Mihon keeps them stable and you have to refresh the randomization by clicking on the random sort option again

themaskedshell:
Thanks for the feedback! I changed the icon to the mui replay icon, and I moved the sort to before the filter. Now the icon matches mihon and the results aren't re-randomized when changing filters. This mostly matches mihons behavior, except here you can repeatedly press the random option to re-randomize the library, while mihon requires you to select another sort mode first.

schroda:
The changelog needs to be updated

CHANGELOG.md
```diff
@@ -46,6 +47,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - (**History**) Show only the last read chapter per manga
 - (**Updates**) Show only the first unread chapter per manga per day
 - (**Manga**) Hide details page title copy button in unsupported environments (secure context (https, localhost) required)
+- (**Sorting/Filtering**) Changed the order from filter->sort to sort->filter
```
schroda:
```suggestion
```

schroda (changes requested):

PR #1136, opened 2026-07-26
feat: show filtered count in library tab and title badges

Seanstoppable:
When filtering a library, I never know how many items are filtered without selecting them all.

This is my attempt at adding this functionality, as well as making sure to highlight similar to the filter icon, to make it more prominent that we are filtering.

### Changes

- When filters or search reduce the visible manga count, the count badge displays `filtered/total` (e.g. `5/20`) instead of just the total
- The count badge uses the same `warning` color as the filter icon when filtering is active, visually tying them together
- Applies to both the Library title badge and the active category tab badge

### DIFF
--- a/src/features/library/screens/Library.tsx
+++ b/src/features/library/screens/Library.tsx
@@ -195,7 +195,8 @@ export function Library() {
             {t`Library`}
             {showTabSize && (
                 <TitleSizeTag
-                    sx={{ ...theme.applyStyles('light', { backgroundColor: 'background.paper' }) }}
+                    sx={{ ...(!isFiltered && theme.applyStyles('light', { backgroundColor: 'background.paper' })) }}
+                    color={isFiltered ? 'warning' : 'default'}
                     label={isFiltered ? `${mangas.length}/${librarySize}` : librarySize}
                 />
             )}
@@ -299,6 +300,7 @@ export function Library() {
                                 {tab.name}
                                 {showTabSize ? (
                                     <TitleSizeTag
+                                        color={tab === activeTab && isFiltered ? 'warning' : 'default'}
                                         label={
                                             tab === activeTab && isFiltered
                                                 ? `${mangas.length}/${tab.mangas.totalCount}`
--- a/src/features/library/screens/Library.tsx
+++ b/src/features/library/screens/Library.tsx
@@ -105,6 +105,7 @@ export function Library() {
         visibleMangas: mangas,
         showFilteredOutMessage,
         filterKey,
+        isFiltered,
     } = useGetVisibleLibraryMangas(categoryMangas, activeTab);
 
     const retryFetchCategoryMangas = useCallback(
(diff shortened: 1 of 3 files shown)


src/features/library/screens/Library.tsx
```diff
@@ -194,13 +195,14 @@ export function Library() {
             {t`Library`}
             {showTabSize && (
                 <TitleSizeTag
-                    sx={{ ...theme.applyStyles('light', { backgroundColor: 'background.paper' }) }}
-                    label={librarySize}
+                    sx={{ ...(!isFiltered && theme.applyStyles('light', { backgroundColor: 'background.paper' })) }}
+                    color={isFiltered ? 'warning' : 'default'}
+                    label={isFiltered ? `${mangas.length}/${librarySize}` : librarySize}
```
schroda:
```suggestion
                    sx={{ ...theme.applyStyles('light', { backgroundColor: 'background.paper' }) }}
                    label={librarySize}
```

src/features/library/screens/Library.tsx
```diff
@@ -194,13 +195,14 @@ export function Library() {
...
                 />
             )}
         </TitleWithSizeTag>,
         t`Library`,
-        [t, showTabSize, librarySize],
+        [t, showTabSize, librarySize, isFiltered, mangas.length],
```
schroda:
```suggestion
        [t, showTabSize, librarySize],
```

src/features/library/screens/Library.tsx
```diff
@@ -296,7 +298,16 @@ export function Library() {
...
+                                {showTabSize ? (
+                                    <TitleSizeTag
+                                        color={tab === activeTab && isFiltered ? 'warning' : 'default'}
+                                        label={
+                                            tab === activeTab && isFiltered
+                                                ? `${mangas.length}/${tab.mangas.totalCount}`
+                                                : tab.mangas.totalCount
+                                        }
+                                    />
+                                ) : null}
```
schroda:
```suggestion
                                {showTabSize ? <TitleSizeTag label={mangas.length} /> : null}
```

I think the highlighted filter icon is enough indication that items are filtered.
The color change feels off and the `x/total` text would maybe be ok now, with only the active category showing the visible item count. If all other categories were showing the filtered counts as well, that would be overloaded and ugly imo

schroda (changes requested):
The changelog needs to be updated accordingly

src/features/library/screens/Library.tsx
Seanstoppable: Sure, no problem.
I keep not noticing the filter icon myself, but happy to drop the coloring for now
Agree that the entire library count updating was probably excessive/annoying.
I DO think showing filter count/total is helpful context. I've simplified how that works. If you still disagree, happy to drop

schroda:
You have to fix the formatting

Seanstoppable:
Oops, fixed now!

PR #1145, opened 2026-08-08
Add hidden library categories with a client-side lock

manilkadev3-max:
## Summary

- allow selected categories to be hidden from the library interface
- reveal hidden categories by entering an access code through Library search
- provide manual locking and an optional configurable timed relock
- store the preference through existing WebUI metadata

## Scope

This feature provides client-side UI concealment for shared browser sessions. It is not an access-control boundary.

An authenticated user with API access or control over the browser can bypass the lock and access the underlying category data.

## Testing

- `pnpm format:check`
- `pnpm tsc`
- `pnpm lint`
- `pnpm build`
- verified category configuration, correct and incorrect access codes, search-field unlocking, manual and automatic relocking, persistence, and desktop and mobile layouts

### DIFF
--- a/src/base/components/AppbarSearch.tsx
+++ b/src/base/components/AppbarSearch.tsx
@@ -20,10 +20,11 @@ import { SearchParam } from '@/base/Base.types.ts';
 
 interface IProps {
     isClosable?: boolean;
+    onSubmit?: (query: string) => boolean | Promise<boolean>;
 }
 
 export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
-    const { isClosable = true } = props;
+    const { isClosable = true, onSubmit } = props;
 
     const theme = useTheme();
     const { t } = useLingui();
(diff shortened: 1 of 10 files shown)


schroda:
Client side password protection makes no sense. Anyone who has access to the server can just modify the client data

manilkadev3-max:
Thank you for pointing this out. You are right that verification and filtering performed entirely by the client cannot provide an access-control boundary. I have updated the title, description, and user-facing wording to describe this as client-side concealment using an access code; it does not restrict server API access.

Would this explicitly limited scope be acceptable for the WebUI? If actual access control is required, would a fully server-backed implementation be within the project scope, with verification, unlock expiration, and category filtering enforced by Suwayomi-Server? Otherwise, I am happy to close this PR and keep the feature as a local customization.

schroda:
I think making it possible to hide categories is enough. The whole "client side authentication" stuff is overkill

PR #1153, opened 2026-08-31
feat: add a sync button to the app bar

kaiserbh:
Adds a cloud-sync icon to the app bar (shown only when sync is enabled on the server) that starts a manual sync on tap and shows a spinner while one is running.

### DIFF
--- a/src/features/navigation-bar/components/DefaultNavBar.tsx
+++ b/src/features/navigation-bar/components/DefaultNavBar.tsx
@@ -24,7 +24,6 @@ import { DesktopSideBar } from '@/features/navigation-bar/components/DesktopSide
 import { useResizeObserver } from '@/base/hooks/useResizeObserver.tsx';
 import { MobileBottomBar } from '@/features/navigation-bar/components/MobileBottomBar.tsx';
 import { useNavBarContext } from '@/features/navigation-bar/NavbarContext.tsx';
-import { SyncButton } from '@/features/sync/components/SyncButton.tsx';
 import { useMetadataServerSettings } from '@/features/settings/services/ServerSettingsMetadata.ts';
 import { NAVIGATION_BAR_ITEMS } from '@/features/navigation-bar/NavigationBar.constants.ts';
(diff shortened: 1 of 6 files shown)


src/features/sync/components/SyncButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+            makeToast(t`Could not start sync`, 'error', getErrorMessage(e));
+        }
+    };
+
+    return (
+        <CustomTooltip title={isSyncing ? t(SYNC_STATE_TRANSLATION[syncState]) : t`Sync now`}>
+            <IconButton onClick={startSync} disabled={isSyncing} color="inherit">
+                {isSyncing ? <CircularProgress size={24} color="inherit" /> : <CloudSyncIcon />}
+            </IconButton>
+        </CustomTooltip>
```
schroda, line 57:
```suggestion
        <CustomTooltip title={isSyncing ? t(SYNC_STATE_TRANSLATION[syncState]) : t`Sync now`}>
            {isSyncing ? (
                <CircularProgress size={24} color="inherit" />
            ) : (
                <IconButton onClick={startSync} disabled={isSyncing} color="inherit">
                    <CloudSyncIcon />
                </IconButton>
            )}
        </CustomTooltip>
```

The tooltip doesn't work on disabled buttons

src/features/navigation-bar/components/DefaultNavBar.tsx
```diff
@@ -166,6 +167,7 @@ export function DefaultNavBar() {
                         >
                             {title}
                         </Typography>
+                        <SyncButton />
```
schroda: I think it's enough to show this button in the library and updates page, same as the library update button

schroda (changes requested):

src/features/sync/components/SyncButton.tsx
kaiserbh: Thanks fixed.

schroda (approved):

PR #1158, opened 2026-09-14
Add fuzzy search and autocomplete to the search bar

Aryan795:
## What

Adds two things to the search bar:

1. **Fuzzy search** for the library — a query now matches even with typos, transposed letters or skipped words, and the closest matches are shown first.
2. **Autocomplete** — the search field suggests previously submitted searches and the titles of the screen being searched.

## Why

Library search is an exact substring match (`enhancedCleanup(...).includes(...)`), so a single typo returns nothing and the user has to retype the query. Manga titles are long, often transliterated, and frequently contain words that are easy to misspell, which makes exact matching a poor fit — "omnisient reader" finds nothing today.

The autocomplete addresses the same problem from the other side: most searches are for something already in the library, so offering those titles avoids typing them in full.

## How it works

`src/base/utils/FuzzySearch.ts` is a self-contained matcher with no new dependency. It returns a score, or `null` for no match, in three bands so match quality stays comparable between candidates:

| kind | score | example against "Tower of God" |
| --- | --- | --- |
| substring | 0.80 – 1.00 | `tower` |
| subsequence | 0.50 – 0.79 | `tog` |
| typo | 0.35 – 0.49 | `twoer` |

- Typos are matched with an **optimal string alignment distance** (Levenshtein plus adjacent transpositions, so `twoer` is one edit from `tower`, not two), with a budget of 1 edit for short queries and 2 for longer ones. Only whole words are compared, so edits cannot match anywhere in a long text.
- Subsequence matches require the matched characters to make up at least half of the span they cover, which keeps `nano` from matching "I'm Not That Kind of Talent". A query that is an **acronym of consecutive words** is exempt, so `tbate` still finds "The Beginning After the End".
- Fields are weighted, with the title weighted highest. The description stays an exact substring match and is weighted below every title match — it is by far the longest field, and matching it fuzzily would both dominate the runtime and match nearly everything.

Autocomplete is a `freeSolo` MUI `Autocomplete` around the existing `SearchTextField`. Submitted searches are kept in local storage (10 most recent), each removable from the dropdown. Suggestions are ranked with the same rule the search itself will use — fuzzy while the setting is on, a plain substring match while it is off — so the dropdown can never offer a title that the search it runs then fails to find.

## Screenshots

Both states are the same library on the same server, with only the **Fuzzy search** setting changed between them.

**Fuzzy search off** — the typo "archmage retruns" offers no suggestion and finds nothing:

1-fuzzy-off-typo-no-suggestions
2-fuzzy-off-typo-no-results

**The setting**, in Library settings:

3-library-settings-fuzzy-toggle

**Fuzzy search on** — the same typo suggests the title and finds the manga:

4-fuzzy-on-typo-suggests-title
5-fuzzy-on-typo-finds-manga

## Behaviour changes to be aware of

- **Fuzzy search defaults to on**, and while a query is active results are ordered by relevance rather than the configured library sort. Both are behind a **"Fuzzy search"** toggle in Library settings; turning it off restores exactly the previous matching and ordering.
- `SearchTextField` now merges `slotProps` instead of replacing them. It previously dropped every slot other than `input`, which silently discarded the props `Autocomplete` passes through `htmlInput` (input ref, keyboard handling). No existing caller relied on the old behaviour.

## Performance

The matcher is allocation-conscious, since a search scores every manga in the library across several fields:

- the edit-distance rows are reused across calls instead of being allocated per comparison
- words are compared through offsets rather than `split(' ')` and substrings
- the query is normalized once per search rather than once per manga and field
- the autocomplete keeps only the best N suggestions while scanning instead of scoring everything into an array and sorting it, and retains nothing between keystrokes
- suggestion ranking is debounced, and skipped entirely while the search bar is closed
- browsing the library without a query skips scoring altogether

Measured over 20 passes of 5,000 titles × 7 queries (700k comparisons): **11.4 ms per pass, with no GC events**.

## Testing

Verified against a real 65-entry library:

| query | result |
| --- | --- |
| `archmage retruns` | The Archmage Returns After 4000 Years |
| `nano machien` | Nano Machine |
| `swordmastr` | The Swordmaster's Son, The Stellar Swordmaster |
| `tbate` | The Beginning After the End |
| `nnmchne` | Nano Machine |
| `nano` | Nano Machine only |

Checked at 1280px and at 390px, including the suggestion dropdown width, touch target sizes, and that the appbar actions are not pushed out of the toolbar.

## Notes

- No GraphQL query changed, so no `versionToServerVersionMapping.json` update is needed.


Aryan795:
Split up as requested and switched to Fuse.js instead of the hand-rolled matcher: #1159 (fuzzy search), #1160 (autocomplete), #1161 (search history). They build on each other in that order.

PR #1159, opened 2026-09-15
Add fuzzy search to the library search

Aryan795:
Part 1 of 3, splitting up #1158 as requested.

## What

Makes the library search typo tolerant, using [Fuse.js](https://www.fusejs.io/), and orders the results by how well they match.

## Why

Library search is an exact substring match (`enhancedCleanup(...).includes(...)`), so a single typo returns nothing and the query has to be retyped. Manga titles are long, often transliterated, and full of words that are easy to misspell — "omnisient reader" finds nothing today.

## How

`src/base/utils/FuzzySearch.ts` holds the shared Fuse configuration so that every typo tolerant search of the app behaves the same:

| option | why |
|---|---|
| `threshold: 0.35` | how wrong a query may be. Tuned against a real library: it still finds "The Archmage Returns After 4000 Years" from "archmage retruns", while unrelated queries keep finding nothing |
| `ignoreLocation` | Fuse otherwise only looks at the start of a text, so a title could not be found by a word in the middle of it |
| `useTokenSearch` + `tokenMatch: 'all'` | every word of the query has to match, but in any order, so "viewpoint omniscient" finds "Omniscient Reader's Viewpoint" |
| `minMatchCharLength: 2` | keeps a single typed character from matching everything |
| `ignoreDiacritics` | accents should not have to be typed |

The library search passes the same fields it already searched as weighted keys, the title weighted highest and the description lowest — it is the longest field by far and therefore the noisiest.

The non-search filters are applied before the index is built, so it only ever covers entries that can actually be a result.

## Screenshots

Same library on the same server, with only the **Fuzzy search** setting changed between them.

**Off** — the typo "archmage retruns" finds nothing:

fuzzy search off, no results

**On** — the same typo finds the entry:

fuzzy search on, finds the manga

**The setting**, in the library settings:

fuzzy search setting

## Behaviour change

Fuzzy search **defaults to on**, and while a query is active the results are ordered by how well they match rather than by the configured library sort. Both are behind a **"Fuzzy search"** toggle in the library settings — turning it off restores exactly the previous matching and ordering. Happy to flip the default if you would rather have it opt-in.

## Testing

Verified against a real 65 entry library:

| query | finds |
|---|---|
| `archmage retruns` | The Archmage Returns After 4000 Years |
| `nano machien` | Nano Machine |
| `omnisient` | Omniscient Reader's Viewpoint |
| `swordmastr` | The Swordmaster's Son, The Stellar Swordmaster |
| `shdow slve` | Shadow Slave |
| `viewpoint omniscient` | Omniscient Reader's Viewpoint |
| `zzzzqqqq` | nothing |

Also checked that the weighted keys work for the array field (`genre`), the nested one (`source.displayName`) and entries where a field is `null`.

Indexing 5,000 entries takes about a millisecond, a search about 9 ms.

## Notes

- No GraphQL query changed, so no `versionToServerVersionMapping.json` update is needed.


Aryan795:
One behaviour change worth flagging: a comma separated query like `action,romance` previously meant "has all of these genres". Fuse treats the comma as part of the query text and matches its words across all searched fields instead, so it now behaves like a normal multi word search. Happy to special case commas to keep the old semantics if you would prefer that.

src/base/utils/FuzzySearch.ts
```diff
@@ -0,0 +1,50 @@
...
+    tokenMatch: 'all',
+} as const satisfies IFuseOptions<unknown>;
+
+/**
+ * Creates a reusable index for "items".
+ *
+ * Building the index is the expensive part of a search - it is cheap enough to be unnoticeable (about a millisecond for
+ * a few thousand entries), but it should still be kept for as long as the items do not change, rather than be rebuilt
+ * for every query.
+ */
```
schroda:
```suggestion

```

src/base/utils/FuzzySearch.ts
```diff
@@ -0,0 +1,50 @@
...
+export const createFuzzySearch = <Item>(items: readonly Item[], keys: FuseOptionKey<Item>[]): Fuse<Item> =>
+    new Fuse([...items], { ...FUZZY_SEARCH_OPTIONS, keys });
+
+/**
+ * Searches a {@link createFuzzySearch} index, returning the matches ordered by how well they match.
+ *
+ * Passing a "limit" lets Fuse keep only that many matches while scoring instead of sorting every match and then
+ * throwing most of them away, which is what a caller that only shows a handful of them wants.
+ */
```
schroda:
```suggestion

```

src/base/utils/FuzzySearch.ts
```diff
@@ -0,0 +1,50 @@
...
+export const fuzzySearch = <Item>(index: Fuse<Item>, query: string, limit?: number): Item[] =>
+    index.search(query, limit != null ? { limit } : undefined).map(({ item }) => item);
```
schroda:
```suggestion
export const fuzzySearch = <Item>(index: Fuse<Item>, ...args: Parameters<Fuse<Item>['search']>): Item[] =>
    index.search(...args).map(({ item }) => item);
```

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -184,20 +185,31 @@ const filterManga = (
     statusFilter(hasStatus, manga) &&
     sourceFilter(hasSource, manga);
 
+/**
+ * Which fields a fuzzy search looks at, and how much each of them says about what the user meant. Searching the library
+ * is almost always searching for a title, while the description is the noisiest field by far, being the longest.
+ */
```
schroda:
```suggestion

```

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -184,20 +185,31 @@ const filterManga = (
...
+const FUZZY_SEARCH_KEYS = [
```
schroda:
```suggestion
const FUZZY_SEARCH_WEIGHTS = [
```

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -184,20 +185,31 @@ const filterManga = (
...
+    { name: 'title', weight: 10 },
+    { name: 'genre', weight: 8 },
+    { name: 'author', weight: 8 },
+    { name: 'artist', weight: 8 },
+    { name: 'source.displayName', weight: 6 },
+    { name: 'sourceId', weight: 6 },
+    { name: 'description', weight: 2 },
```
schroda:
```suggestion
    { name: 'title', weight: 10 },
    { name: 'author', weight: 4 },
    { name: 'artist', weight: 4 },
    { name: 'genre', weight: 2.5 },
    { name: 'source.displayName', weight: 1 },
    { name: 'sourceId', weight: 1 },
    { name: 'description', weight: 0.5 },
```

The title should have by far the most weight and I think that the author/artist should have more weight than a list of genre which are relatively short words

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -184,20 +185,31 @@ const filterManga = (
...
+];
+
 type TMangasFilter = TMangaQueryFilter & TMangaFilter;
+
+/** Applies everything but the search, so that a fuzzy index only ever covers entries that can actually be a result. */
```
schroda:
```suggestion

```

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -364,6 +379,21 @@ export const useGetVisibleLibraryMangas = <Manga extends MangaIdInfo & TMangasFi
         ],
     );
 
+    // building the index is the expensive part of a fuzzy search, so it is kept for as long as the entries it covers
+    // do not change, rather than being rebuilt for every keystroke
```
schroda:
```suggestion

```

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -374,12 +404,14 @@ export const useGetVisibleLibraryMangas = <Manga extends MangaIdInfo & TMangasFi
...
+        visibleMangas.length === 0 &&
         mangas.length > 0;
 
     return {
-        visibleMangas: filteredMangas,
+        visibleMangas,
         showFilteredOutMessage,
-        filterKey: `${JSON.stringify(options)}${settings.ignoreFilters}`,
+        // the query and the source filter belong in here as well: both change which entries are shown, and with
+        // fuzzy search they also change their order, which is what the remount works around (react-virtuoso#1242)
```
schroda:
```suggestion

```

src/features/library/screens/LibrarySettings.tsx
```diff
@@ -179,6 +179,17 @@ export function LibrarySettings() {
                         onChange={(e) => setSettingValue('ignoreFilters', e.target.checked)}
                     />
                 </ListItem>
+                <ListItem>
```
schroda: Why do you think that this should be toggleable?

schroda (changes requested):
> One behaviour change worth flagging: a comma separated query like action,romance previously meant "has all of these genres". Fuse treats the comma as part of the query text and matches its words across all searched fields instead, so it now behaves like a normal multi word search. Happy to special case commas to keep the old semantics if you would prefer that.

I don't think that that's a problem

src/features/library/hooks/useGetVisibleLibraryMangas.ts
```diff
@@ -364,6 +379,21 @@ export const useGetVisibleLibraryMangas = <Manga extends MangaIdInfo & TMangasFi
...
+    // building the index is the expensive part of a fuzzy search, so it is kept for as long as the entries it covers
+    // do not change, rather than being rebuilt for every keystroke
+    const fuzzySearchIndex = useMemo(
+        () => (isFuzzySearchActive ? createFuzzySearch(filteredMangas, FUZZY_SEARCH_KEYS) : null),
+        [isFuzzySearchActive, filteredMangas],
+    );
+
+    const visibleMangas = useMemo(() => {
+        if (fuzzySearchIndex) {
+            return fuzzySearch(fuzzySearchIndex, query!);
```
schroda:
```suggestion
        if (fuzzySearchIndex && query) {
            return fuzzySearch(fuzzySearchIndex, query);
```

Aryan795:
Applied all of it, thanks — weights adjusted as suggested, comments dropped, `fuzzySearch` now forwards Fuse's own search arguments, and the non-null assertion is gone. #1160 and #1161 are rebased on top with the call sites updated.

Aryan795:
Applied all of it, thanks — weights adjusted as suggested, comments dropped, `fuzzySearch` now forwards Fuse's own search arguments, and the non-null assertion is gone. #1160 and #1161 are rebased on top with the call sites updated.

schroda (approved):

PR #1160, opened 2026-09-15
Add autocomplete to the search bar

Aryan795:
Part 2 of 3, splitting up #1158 as requested. Follows #1159.

**Depends on #1159** — it reuses the shared matcher added there, so that the suggestions behave the same way the search does. Please merge that one first; this branch is based on it and the diff will shrink to its own four files once it lands.

## What

The search bar suggests the entries of the screen that is being searched while typing.

## Why

Most searches are for something already on the screen being searched, so offering those entries avoids typing a long title out in full.

## How

`AppbarSearch` now renders a `freeSolo` MUI `Autocomplete` around the existing `SearchTextField`. The library screen passes its titles; the other screens keep the plain input until they have something worth suggesting.

Suggestions are matched exactly the way the search itself matches — fuzzily while the **Fuzzy search** setting is on, as a plain substring while it is off — so the dropdown can never offer an entry that the search it then runs fails to find.

Ranking is debounced and skipped entirely while the search bar is closed.

## Screenshots

Typing the typo "archmage retruns" suggests the entry it is looking for:

autocomplete suggesting the title

With **Fuzzy search** turned off the same typo offers nothing, because the search it would run finds nothing either:

no suggestion with fuzzy search off

## Included fix

`SearchTextField` merged into this: it replaced the whole `slotProps` object instead of merging it, which silently dropped every slot other than `input`. That is where `Autocomplete` passes the props wiring up its input — the ref and the keyboard handling — so the field would not work inside one. No existing caller relied on the old behaviour.

## Testing

Verified against a real library at 1280px and at 390px: the dropdown width, that long titles are cut off rather than stretching the row, and that the appbar actions are not pushed out of the toolbar on a narrow screen.

`pnpm lint`, `pnpm format:check`, `pnpm tsc` and `pnpm build` all pass.

PR #1161, opened 2026-09-15
Remember submitted searches in the search bar

Aryan795:
Part 3 of 3, splitting up #1158 as requested. Follows #1159 and #1160.

**Depends on #1160** (which in turn depends on #1159) — the history is shown in the dropdown that one adds. Please merge those first; this branch is based on them and the diff will shrink to its own files once they land.

## What

The search bar remembers the queries that were actually submitted and offers them again.

## Why

Searches repeat, both on the same screen and across screens — the history is shared by every search input of the app, so a search done in the library can be repeated in a source without typing it again.

## How

The ten most recent queries are kept in local storage, most recent first, de-duplicated case insensitively. They are shown under their own **Recent searches** group in the dropdown, and each one can be removed individually.

`openOnFocus` is set because an empty input otherwise keeps the popup closed, which would make the previous searches — the only thing an empty search box has to offer — unreachable.

## Testing

Verified against a real library that submitting a query records it, that it is offered again on focusing the empty search bar, and that removing an entry removes it without running the search instead.

On a 390px viewport the remove button follows the 44px touch target guideline, since missing it would otherwise run the search rather than remove it.

## Notes


src/base/components/AppbarSearch.tsx
```diff
@@ -64,6 +75,8 @@ export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
         settings: { fuzzySearch: isFuzzySearchEnabled },
     } = useMetadataServerSettings();
 
+    const { history, addToHistory, removeFromHistory } = useSearchHistory();
```
schroda: This needs to be scoped by "search origin". I.e., library searches should not show up in other places like the source browse or browse extensions tab

src/base/components/AppbarSearch.tsx
```diff
@@ -82,24 +95,54 @@ export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
...
+    // surfaces the previous search it is repeating
+    const historySearchIndex = useMemo(
+        () => (isOpen && isFuzzySearchEnabled ? createFuzzySearch(history, []) : null),
+        [isOpen, isFuzzySearchEnabled, history],
+    );
 
-        // the suggestions of a closed search box are never shown, and without anything typed they would just be the
-        // unsorted content of the screen, which says nothing
-        if (!isOpen || !trimmedSearchString) {
+    const options = useMemo<SearchSuggestion[]>(() => {
```
schroda, line 98: I'd say the history should only be shown in case nothing is typed. Otherwise, it's replaced by the suggestions

schroda (changes requested):

Aryan795:
Both addressed. The history is now scoped by a required `searchOrigin` prop (library / source / global search / extensions), each stored under its own key, and it is only shown while nothing is typed — suggestions replace it as soon as there is a query, which also removed the need to match the history at all.

src/base/hooks/useSearchHistory.ts
```diff
@@ -0,0 +1,74 @@
...
+const MAX_ENTRY_LENGTH = 100;
+
+const areEntriesEqual = (a: string, b: string): boolean => baseCleanup(a) === baseCleanup(b);
+
+/**
+ * The queries that were actually submitted for a given "origin", most recent first, so that the searches of e.g. the
+ * library stay separate from the ones of the source browse.
+ */
+export const useSearchHistory = (
+    origin: SearchOrigin,
```
schroda:
```suggestion
    origin: SearchOrigin,
    maxSize = 10,
```

instead of the `MAX_ENTRIES` const.
Then `MAX_HISTORY_SUGGESTIONS` only has to be passed without having to handle the length additionally

schroda:
```suggestion
   key: string,
```

src/base/Base.types.ts
```diff
@@ -43,3 +43,10 @@ export enum SearchParam {
     QUERY = 'query',
     REDIRECT = 'redirect',
 }
+
+export enum SearchOrigin {
+    LIBRARY = 'library',
+    SOURCE = 'source',
+    GLOBAL_SEARCH = 'globalSearch',
+    EXTENSIONS = 'extensions',
+}
```
schroda:
```suggestion
```

An enum for that is too much

src/base/hooks/useSearchHistory.ts
```diff
@@ -0,0 +1,74 @@
...
+        [storedHistory],
+    );
+
+    const addToHistory = useCallback(
+        (query: string) => {
+            const trimmedQuery = query.trim();
+
+            if (!trimmedQuery || trimmedQuery.length > MAX_ENTRY_LENGTH) {
+                return;
+            }
```
schroda:
```suggestion
            const trimmedQuery = query.slice(0, MAX_ENTRY_LENGTH).trim()
```
The value shouldn't just get omitted

src/base/components/AppbarSearch.tsx
```diff
@@ -157,6 +179,8 @@ export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
                 options={options}
                 // the options are already ranked by relevance, re-filtering them would drop the typo tolerant hits
                 filterOptions={(unfilteredOptions) => unfilteredOptions}
+                getOptionLabel={(option) => (typeof option === 'string' ? option : option.label)}
+                groupBy={(option) => (option.isFromHistory ? t`Recent searches` : t`Suggestions`)}
```
schroda:
```suggestion
                groupBy={(option) => (option.isFromHistory ? t`Recent searches` : '')}
```

src/base/components/AppbarSearch.tsx
```diff
@@ -166,12 +190,29 @@ export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
...
+                    handleChange(typeof value === 'string' ? value : value.label);
                 }}
                 renderOption={({ key, ...optionProps }, option) => (
                     <Box key={key} component="li" sx={{ gap: 1 }} {...optionProps}>
-                        <SearchIcon fontSize="small" />
-                        <TypographyMaxLines>{option}</TypographyMaxLines>
+                        {option.isFromHistory ? <HistoryIcon fontSize="small" /> : <SearchIcon fontSize="small" />}
+                        <TypographyMaxLines sx={{ flexGrow: 1 }}>{option.label}</TypographyMaxLines>
+                        {option.isFromHistory && (
+                            <CustomTooltip title={t`Remove from search history`}>
```
schroda:
```suggestion
                            <CustomTooltip title={t`Delete`}>
```

schroda (changes requested):

src/base/hooks/useSearchHistory.ts
```diff
@@ -0,0 +1,74 @@
...
+        `${SEARCH_HISTORY_STORAGE_KEY}::${origin}`,
+        STABLE_EMPTY_ARRAY,
+    );
+
+    // the stored value is whatever happens to be under the key, so anything but a list of strings is ignored rather
+    // than being handed to the caller, where it would throw
+    const history = useMemo(
+        () => (Array.isArray(storedHistory) ? storedHistory.filter((entry) => typeof entry === 'string') : []),
+        [storedHistory],
+    );
```
schroda:
```suggestion
    const [history, setHistory] = useLocalStorage<string[]>(
        `${SEARCH_HISTORY_STORAGE_KEY}::${origin}`,
        STABLE_EMPTY_ARRAY,
    );
```

if someone messes with the values it's their own fault

src/base/components/AppbarSearch.tsx
```diff
@@ -89,33 +155,93 @@ export const AppbarSearch: React.FunctionComponent<IProps> = (props) => {
...
+                                <IconButton
+                                    size="small"
+                                    edge="end"
+                                    aria-label={t`Remove from search history`}
+                                    sx={{ width: { xs: 44, sm: 32 }, height: { xs: 44, sm: 32 } }}
+                                    onMouseDown={(e) => e.preventDefault()}
+                                    onClick={(e) => {
+                                        e.stopPropagation();
+                                        removeFromHistory(option.label);
+                                    }}
```
schroda:
```suggestion
                                    edge="end"
                                    onClick={(e) => {
                                        e.stopPropagation();
                                        removeFromHistory(option.label);
                                    }}
```

schroda:
You have to rebase onto master

Aryan795:
Rebased onto master and applied all of it — `useSearchHistory(key, maxSize)` with no enum, over-long queries truncated instead of dropped, the array guard dropped, only the history group labelled, and the delete button trimmed down.

src/base/components/AppbarSearch.tsx
```diff
@@ -34,16 +37,25 @@ const MAX_SUGGESTIONS = 8;
...
+type SearchSuggestion = {
+    label: string;
+    isFromHistory: boolean;
+};
+
 const getSubstringMatches = (query: string, suggestions: string[]): string[] =>
     suggestions.filter((suggestion) => enhancedCleanup(suggestion).includes(query));
 
 interface IProps {
+    /** keeps the search history of e.g. the library separate from the one of the source browse */
```
schroda, line 51:
```suggestion
```

src/base/hooks/useSearchHistory.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/** Long enough for any realistic search, short enough to not fill up the local storage with a pasted wall of text. */
+const MAX_ENTRY_LENGTH = 100;
+
+const areEntriesEqual = (a: string, b: string): boolean => baseCleanup(a) === baseCleanup(b);
+
+/**
+ * The queries that were actually submitted for a given "key", most recent first, so that the searches of e.g. the
+ * library stay separate from the ones of the source browse.
+ */
```
schroda, line 24:
```suggestion
```

CHANGELOG.md
schroda, line 11:
```suggestion
- (**Global/Search**) Increase appbar search width when active
- (**Global/Search**) Remember submitted searches and offer them in the search bar, each removable from the dropdown
```

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - (**Library**) Add autocomplete to the search bar, suggesting the entries of the category that is being searched. Pressing `tab` selects the suggested live auto-completion
 - (**Sync**) Add sync button to the app bar of the library and updates pages
 - (**Browse/Extension**) Add search suggestions to extensions browse tab
+- (**Global/Search**) Remember submitted searches and offer them in the search bar, each removable from the dropdown
```
schroda, line 21:
```suggestion
```

schroda (changes requested):

Aryan795:
Applied all four suggestions — removed both comments and moved the changelog entry next to the other `Global/Search` line.

schroda (approved):

